PR #2, opened 2021-03-30
Create checksums in ISO

dbnicholson:
The idea here is to create `.sha256` files in the ISO so that eos-installer can use them for verification when signatures aren't available. See endlessm/eos-installer#91.

I'm not sure if it would be better to have a `SHA256SUMS` file as would be normally produced by `sha256sum` but it seemed a little odd because the image file is actually embedded in the squashfs image, so you couldn't actually do something like run `sha256sum -c`.

https://phabricator.endlessm.com/T31304


dbnicholson:
This was superseded by #5.

PR #3, opened 2021-03-30
No default signing and publishing

dbnicholson:
Have the builder not sign or publish anything by default. This way someone can simply run `./eos-image-builder` and it succeed without any special privileges. The settings needed for Endless builds have moved to our private config repo, so there should be no changes in our case. While here cleanup some old handling of the output directory.

https://phabricator.endlessm.com/T31698


run-build
```diff
@@ -581,12 +581,8 @@ class ImageBuilder(object):
             cmd = (chrootscript, self.builddir, buildscript)
             subprocess.check_call(cmd, env=builder_env)
 
-        # Unless this is a dry run, the output files have been published
-        # and can be deleted.
-        if not self.dry_run:
-            outrootdir = self.get_build_option('outrootdir')
-            log.info('Cleaning up output in %s', outrootdir)
-            shutil.rmtree(outrootdir, ignore_errors=True)
+        log.info('Built assets available in %s',
```
wjt, line 584: This is a really nice change. I had definitely been in situations when I've run this locally and it's cheerfully deleted my image at the end of the run :)

wjt (approved):

PR #8, opened 2021-04-05
Full sha256sum files

dbnicholson:
Include the target in the individual checksum files so that they can be passed to `sha256sum --check` like the combined checksum file.

https://phabricator.endlessm.com/T31076

### DIFF
--- a/helpers/split-image
+++ b/helpers/split-image
@@ -182,8 +182,8 @@ split_image() {
   disk2_img_asc=$(eib_outfile disk2.img.asc)
 
   local disk1_img_csum disk2_img_csum
-  disk1_img_asc=$(eib_outfile disk1.img.sha256)
-  disk2_img_asc=$(eib_outfile disk2.img.sha256)
+  disk1_img_csum=$(eib_outfile disk1.img.sha256)
+  disk2_img_csum=$(eib_outfile disk2.img.sha256)
 
   # Mount root filesystem from the last partition of the image file.
   DISK1_LOOP=$(losetup --show -f "${disk1_img}")
--- a/helpers/split-image
+++ b/helpers/split-image
@@ -181,9 +181,11 @@ split_image() {
   disk1_img_asc=$(eib_outfile disk1.img.asc)
   disk2_img_asc=$(eib_outfile disk2.img.asc)
 
-  local disk1_img_csum disk2_img_csum
+  local disk1_img_csum disk1_img_csum_tgt disk2_img_csum disk2_img_csum_tgt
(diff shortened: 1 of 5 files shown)


helpers/split-image
```diff
@@ -182,8 +182,8 @@ split_image() {
   disk2_img_asc=$(eib_outfile disk2.img.asc)
 
   local disk1_img_csum disk2_img_csum
-  disk1_img_asc=$(eib_outfile disk1.img.sha256)
```
wjt, line 185: This is Right and Proper, though split image support is going away https://phabricator.endlessm.com/T30922 so it's a little moot.

lib/eib.sh
```diff
@@ -271,12 +271,14 @@ sign_file() {
...
-# Create a detached checksum with sha256sum.
+# Create a detached checksum with sha256sum. By default the target in
+# the checksum file is the basename of the file being checksummed.
 checksum_file() {
   local file=${1:?No file supplied to ${FUNCNAME}}
   local outfile=${2:-${file}.sha256}
+  local target=${3:-${file##*/}}
 
-  sha256sum "${file}" | cut -d' ' -f1 > "${outfile}"
+  sha256sum "${file}" | awk "{print \$1 \"  $target\"}" > "${outfile}"
```
wjt, line 281: I don't quite understand why the `awk` step is needed – to strip the path down to a basename? Would another option be to `cd` to the directory the target is in in a subshell? Maybe it's no easier.

dbnicholson: Because the path you're using for the checksum is not necessarily the same as what you want in the file. It could be because you're not in the same directory or it could be because the file name doesn't match what you want at the time the command is run. That's what happens in the extracted split case and could be in the extracted image case, although the raw image temporary file is named like the output file.

dbnicholson: Would it be easier to read if the output from sha256sum was captured and the file was written manually? To me it seems like changing directories won't cover the case where the file isn't named in the expected way yet and it's not hard to tell the function what you want to have in there.

wjt: I'm not necessarily worried about its legibility, more about why it's there. It's probably fine like thi.s

dbnicholson: To me it makes sense - what's being generated includes a strong reference to something else, so I'd rather be explicit about what that reference is instead of running the program in just the right way to produce the desired reference. If `sha256sum` offered a way to do that, I'd use that.

I think it would be less icky as a capture and write, though, so I think I'll do that instead of `awk`.

wjt (approved):

PR #9, opened 2021-04-07
Add new default product eoscustom

dbnicholson:
With the image builder now public, presumably people will want to build
their own images. In order to not confuse these images with official
Endless eos images, add a new default product named eoscustom.

This is nearly the same as an eos image with a few exceptions:

* No ISO or VM images

* No automatic Google Chrome

* No xz compression

These settings could be added, but they can also be customized by users
using the normal methods.

https://phabricator.endlessm.com/T31707

### DIFF
--- a/config/product/eoscustom.ini
+++ b/config/product/eoscustom.ini
@@ -1,0 +1,5 @@
+# Settings for the eoscustom product
+
+[ostree]
+# Use the standard eos ostree
+product = eos
--- a/lib/eib.py
+++ b/lib/eib.py
@@ -339,7 +339,7 @@ def add_cli_options(argparser):
 
     argparser.add_argument('-d', '--localdir', help='local settings directory')
 
-    add_argument('-p', '--product', default='eos',
+    add_argument('-p', '--product', default='eoscustom',
                  help='product to build')
     argparser.add_argument('-a', '--arch', choices=SUPPORTED_ARCHES,
                            help='architecture to build '
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -32,7 +32,7 @@ def builder_config(config):
     interpolation of full sections should succeed.
     """
     config[config.BUILD_SECTION].update({
-        'product': 'eos',
+        'product': 'eoscustom',
         'branch': 'master',
         'arch': 'amd64',
         'platform': 'amd64',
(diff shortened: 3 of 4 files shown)


dbnicholson:
I was torn on whether to maintain the minor changes that the `eos` product has from the defaults or not. I decided to leave it as close to the defaults as possible, which also makes it very minimal with the `base` personality. I was able to build an image in 8 minutes from that, which is nice.

wjt:
I predict that JP will say we want ISOs by default and I tend to agree, though I know it will significantly increase the build time. I agree with the other choices.

> I was able to build an image in 8 minutes from that, which is nice.

Nice! Before this change, with a warm cache, the default build (xz, iso, VM images) took 40 minutes. Running this branch now to compare...

jprvita:
> I predict that JP will say we want ISOs by default and I tend to agree, though I know it will significantly increase the build time. I agree with the other choices.
> 

I was on the fence about this -- I do believe in ISOs as *our* format for image distribution, but that does not necessarily extends to images built by 3rd parties. Now that I gave it a bit of thought I actually agree with @dbnicholson's approach here of disabling ISOs by default for `eoscustom`. I believe in most cases the user will need at least a few iterations in the build-test-tweak loop, so making the build process as quick as possible seems like the way to go IMO.

On that note, (as a follow-up task) I would even go a bit further and add a way to completely disable compression, to save the compression time during build and the extraction time immediately after. When I build things locally I normally neuter the compression routine so I can `./eos-image-builder && qemu-system-x86_64 file.img`.

> > I was able to build an image in 8 minutes from that, which is nice.
> 
> Nice! Before this change, with a warm cache, the default build (xz, iso, VM images) took 40 minutes. Running this branch now to compare... 13 mins.

\o/

dbnicholson:
> On that note, (as a follow-up task) I would even go a bit further and add a way to completely disable compression, to save the compression time during build and the extraction time immediately after. When I build things locally I normally neuter the compression routine so I can `./eos-image-builder && qemu-system-x86_64 file.img`.

I was just mentioning to @wjt that you wanted that. I agree it would be nice to have. When I'm hacking on the image builder the last thing I want to be doing is spending several minutes compressing something that I'm either going to immediately decompress or just throw away entirely. I'm worried that it could be messy, but if you file a task I'll take a look.

PR #14, opened 2021-04-20
hooks: Support installing a subset of a kolibri channel

dylanmccall:
https://phabricator.endlessm.com/T31788

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -202,22 +202,6 @@ kalite_content_user = uploader
 kalite_content_variant = endless
 kalite_content_version = 0.16
 
-# Preloaded Kolibri content. eos-kolibri-help is always selected by default.
-kolibri_content = false
-kolibri_content_app_version = 0.14.6
-kolibri_content_app_desktop_xdg_plugin_version = 1.0.8
-kolibri_content_channel_eos_help = e8a879742b2249a0a4b890f9903916f7
-kolibri_content_channel_endless_key = d29581bbf4394456971b8cb807219c39
-kolibri_content_extra_channels =
-
-# Kolibri automatic provisioning configuration
-kolibri_automatic_provision = true
-kolibri_automatic_provision_facility_name = Endless
-kolibri_automatic_provision_superuser_name =
-kolibri_automatic_provision_superuser_password =
-kolibri_automatic_provision_preset = nonformal
-kolibri_automatic_provision_landing_page = learn
-
 # Product name for ISO labels, etc
 product_name = Endless OS
 
@@ -265,6 +249,43 @@ branding_fbe_config = ${build:datadir}/branding/gnome-initial-setup/default/gnom
 
 # Environment variables to substitute in branding configuration files
 branding_subst_vars_add =
+
+[kolibri]
+app_version = 0.14.6
+app_desktop_xdg_plugin_version = 1.0.8
+
+automatic_provision = true
+automatic_provision_facility_name = Endless
+automatic_provision_superuser_name =
+automatic_provision_superuser_password =
+automatic_provision_preset = nonformal
+automatic_provision_landing_page = learn
+
+# Preloaded Kolibri channels
+# install_channels =
+#   # How to get started with Kolibri on Endless OS
+#   e8a879742b2249a0a4b890f9903916f7
+
+# Kolibri content configuration
+#
+# Each kolibri-<channel> section corresponds to a channel listed in the
+# install_channels option in the kolibri section. These sections are
+# optional. By default, all content from a channel will be included.
+#
+# include_node_ids - List of content node IDs to include. If this is not set,
+#   all content from the channel will be included.
+# exclude_node_ids - A list of content node IDs to exclude. Use this to
+#   exclude content nested beneath content nodes that are being included.
+#
+# Example:
+#
+# [kolibri-e8a879742b2249a0a4b890f9903916f7]
+# include_node_ids =
+#   # English [topic - 23]
+#   3b909a18242c48208dbc49d06bc48162
+# exclude_node_ids =
+#   # Kolibri 0.12.2 User Guide for Admins [document]
+#   5bb37c1832c8489ab2940f31588305f6
 
 [usb]
 size = 16000000000
--- a/hooks/image/60-kolibri-content
+++ b/hooks/image/60-kolibri-content
@@ -1,8 +1,29 @@
 # Populate the Kolibri home directory
 
-if [ "${EIB_IMAGE_KOLIBRI_CONTENT}" != "true" ]; then
+if [ -z "${EIB_KOLIBRI_INSTALL_CHANNELS}" ]; then
   exit 0
 fi
+
+import_kolibri_channel()
+{
+  local channel_id=$1
+  local channel_include_node_ids_var="EIB_KOLIBRI_${channel_id}_INCLUDE_NODE_IDS"
+  local channel_exclude_node_ids_var="EIB_KOLIBRI_${channel_id}_EXCLUDE_NODE_IDS"
+  local importcontent_opts=()
+
+  if [ -n "${!channel_include_node_ids_var}" ]; then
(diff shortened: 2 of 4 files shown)


dylanmccall:
This will also require some significant changes to [endless-image-config](https://github.com/endlessm/endless-image-config). Specifically, https://github.com/endlessm/endless-image-config/blob/master/config/personality/en_US_payg.ini specifies channels the old way (with content magically selected based on `endless_key_resources_channel_id` in https://github.com/endlessm/endless-image-config/blob/master/config/defaults.ini). Instead, we'll need to put a list of selected content nodes for each channel somewhere in the image configuration.

wjt:
I think it would be useful to prepare that endless-image-config PR before merging this – it would provide a good reference for what the desired config syntax is.

jprvita:
> I think it would be useful to prepare that endless-image-config PR before merging this – it would provide a good reference for what the desired config syntax is.

I agree with @wjt here.

> This will also require some significant changes to [endless-image-config](https://github.com/endlessm/endless-image-config). Specifically, [en_US_payg.ini](https://github.com/endlessm/endless-image-config/blob/master/config/personality/en_US_payg.ini) specifies channels the old way (with content magically selected based on `endless_key_resources_channel_id` in [defaults.ini](https://github.com/endlessm/endless-image-config/blob/master/config/defaults.ini)). Instead, we'll need to put a list of selected content nodes for each channel somewhere in the image configuration. I'm not sure if we should merge this before or after doing that :)

The selection for `en_US_payg` at the time was simply "everything that is shipped in the EK". So I think we could simply drop the extra magical channel from there, but at some point we should probably have someone review the content selection for `en_US_payg`.

hooks/image/60-kolibri-content
```diff
@@ -4,6 +4,29 @@ if [ "${EIB_IMAGE_KOLIBRI_CONTENT}" != "true" ]; then
   exit 0
 fi
 
+import_kolibri_channel()
+{
+  local channel_id=$1
+  local channel_include_node_ids_var="EIB_KOLIBRI_${channel_id}_INCLUDE_NODE_IDS"
+  local channel_exclude_node_ids_var="EIB_KOLIBRI_${channel_id}_EXCLUDE_NODE_IDS"
```
jprvita, line 11: If I'm reading this right, we expect to have a section like this in the config, when selecting subsets of channels (here just `<node_id1>` and `<node_id2>` of `<channel_id1>`, and all but `<node_id5>` from `<channel_id3>`):
```
[kolibri]
install_channels = <channel_id1>
                   <channel_id2>
                   <channel_id3>
<channel_id1>_include_node_ids = <node_id1>
                                 <node_id2>
<channel_id3>_exclude_node_ids = <node_id5>
```
Can we add an example in `default.ini`? Even if just something like this in the comments.

config/defaults.ini
```diff
@@ -266,6 +261,13 @@ branding_fbe_config = ${build:datadir}/branding/gnome-initial-setup/default/gnom
 # Environment variables to substitute in branding configuration files
 branding_subst_vars_add =
 
+[kolibri]
+app_version = 0.14.6
+app_desktop_xdg_plugin_version = 1.0.8
```
jprvita, line 266: It would be nice if we could get rid of these hardcoded versions and just use the same flatpak that is being installed in the image, but I think you have tried this in the past and hit some problems, so don't worry too much about it.

jprvita (changes requested):
One small request to document this mechanism in `defaults.ini`, but other than that this looks great! Thanks for taking this on, @dylanmccall!

hooks/image/60-kolibri-content
dbnicholson: This might be reaching the limits of what's reasonably expressable in INI format. Maybe it would be preferable to define the channel selection in a JSON or YAML file in `data/` with the path defined here. Here's some barebones YAML where the top level is a dictionary where keys are the channel IDs and the values are options dictionaries:
```
channel1:
channel2:
  include: node1
channel3:
  exclude: node2
```
That would be pretty easy to parse in python:
```
import yaml
with open('channels.yaml') as f:
    channels = yaml.safe_load(f)
for channel, options in channels.items():
    # Turn None values to empty dicts for simplicity
    options = options or {}
    include_nodes = options.get('include', [])
    exclude_nodes = options.get('exclude', [])
    import_content_cmd = ['kolibri', 'manage', 'importcontent']
    if include_nodes:
        import_content_cmd.append('--node_ids=' + ','.join(include_nodes))
    if exclude_nodes:
        import_content_cmd.append('--exclude_node_ids' + ','.join(include_nodes))
```
And on. In that case I guess I'd split out the importing part of the hook from the kolibri setup part.

dbnicholson (commented):
I think this is probably workable the way you're doing it, but I suspect that the configuration will be pretty awkward strictly in the INI configuration.

dbnicholson:
After discussion in the deployment meeting we decided that in the interest of time, the way you have it is fine and maybe we'll iterate on this afterwards. I would really like to see some documentation and example sections.

wjt:
I think there should be an example in this (open-source) repo; but those who have access can see a practical example at https://github.com/endlessm/endless-image-config/pull/988

dbnicholson:
That's helpful. Although an example would be really nice, what I really want is a description of the sections and options like the description of the `flatpak-remote` sections in `config/defaults.ini`. Someday maybe we'd have an actual documentation format for the configuration schema, but for now comments in the default configuration is the best we've got.

hooks/image/60-kolibri-content
```diff
@@ -4,6 +4,29 @@ if [ "${EIB_IMAGE_KOLIBRI_CONTENT}" != "true" ]; then
...
+  local importcontent_opts=()
+
+  if [ -n "${!channel_include_node_ids_var}" ]; then
+    importcontent_opts+=("--node_ids='$(
+      echo "${!channel_include_node_ids_var}" | tr ' ' ','
+    )'")
+  fi
```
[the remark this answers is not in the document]
dylanmccall: Thanks for the suggestion :) Should be better now.

dylanmccall:
Okay, I added some documentation to defaults.ini in a recent push. Indeed, the configuration gets pretty hairy (especially parsing it), although it turns out surprisingly not terrible looking in practice :b Would be nice to be able to use a more portable means of specifying Kolibri content in the future. That is, if it _is_ a json file, a json file that is generated by and consumed by Kolibri itself.

config/defaults.ini
```diff
@@ -202,13 +202,8 @@ kalite_content_user = uploader
 kalite_content_variant = endless
 kalite_content_version = 0.16
 
-# Preloaded Kolibri content. eos-kolibri-help is always selected by default.
+# Set to true to enable preloaded Kolibri content
 kolibri_content = false
```
jprvita, line 206: Perhaps we should get rid of this `kolibri_content` variable too, and check whether `EIB_KOLIBRI_INSTALL_CHANNELS` is empty or not to decide if we should bail out early in the hooks.

config/defaults.ini
```diff
@@ -202,13 +202,8 @@ kalite_content_user = uploader
...
-kolibri_content_app_version = 0.14.6
-kolibri_content_app_desktop_xdg_plugin_version = 1.0.8
-kolibri_content_channel_eos_help = e8a879742b2249a0a4b890f9903916f7
-kolibri_content_channel_endless_key = d29581bbf4394456971b8cb807219c39
-kolibri_content_extra_channels =
 
 # Kolibri automatic provisioning configuration
```
jprvita, line 208: Maybe we should also move these automatic provisioning settings to the `[kolibri]` section? (feel free to do this as a follow-up work)

jprvita (changes requested):
Looking pretty good! Left a couple of small suggestion for improving it, but I would not oppose to merging as-is.

config/defaults.ini
dylanmccall: Yeah, makes sense. Done :)

PR #15, opened 2021-04-23
tools: Add a tool to generate image configuration for Kolibri

dylanmccall:
https://phabricator.endlessm.com/T31824

I'm not sure where to put this kind of script, so I stuck it in a `tools` directory for now…


jprvita:
> I'm not sure where to put this kind of script, so I stuck it in a `tools` directory for now…

I think we should ship this in the ostree, or (as @wjt suggested) perhaps in the Kolibri flatpak, so it is readily available for partners and the Outreach team to use when curating content for custom images (which means this PR would live somewhere else).

wjt:
The script uses Kolibri's Python API, so on Endless OS at least it must run within the context of the Kolibri Flatpak. Perhaps an additional Django management command, invoked via `eos-kolibri-manage` on Endless OS? I don't think the ostree is the right home for it, because it will need to be kept in sync with Kolibri's internal data model, etc.

(I could imagine situations where partners are using Kolibri on non-Endless OS platforms.)

dylanmccall:
Thanks for the feedback! I'm moving this over to [flathub/org.learningequality.Kolibri](https://github.com/flathub/org.learningequality.Kolibri), along with a change in eos-kolibri-manage that will allow us to invoke it through there. I'll link that PR to this one once it's up…

jprvita:
These changes have been integrated via https://github.com/flathub/org.learningequality.Kolibri/pull/32, so I'm deleting the branch here.

PR #17, opened 2021-04-27
config: Add Kolibri content to the base personality

dylanmccall:
This adds some simple Kolibri content (and automatic provisioning) to the `base` personality.

https://phabricator.endlessm.com/T31866

### DIFF
--- a/config/personality/base.ini
+++ b/config/personality/base.ini
@@ -12,3 +12,9 @@ apps_del = ${apps_add_defaults}
 [flatpak-remote-flathub]
 # Delete all default additions and explicitly build the list from scratch.
 apps_del = ${apps_add_defaults}
+apps_add =
+  org.learningequality.Kolibri
+
+[kolibri]
+install_channels =
+  e8a879742b2249a0a4b890f9903916f7
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -281,11 +281,16 @@ automatic_provision_landing_page = learn
 #
 # [kolibri-e8a879742b2249a0a4b890f9903916f7]
 # include_node_ids =
-#   # English [topic - 23]
+#   # English [topic]
 #   3b909a18242c48208dbc49d06bc48162
 # exclude_node_ids =
 #   # Kolibri 0.12.2 User Guide for Admins [document]
 #   5bb37c1832c8489ab2940f31588305f6
+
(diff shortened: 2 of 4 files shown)


dylanmccall:
Alternatively, we could set `automatic_provision = false` here. Keeping it is consistent with other images that include Kolibri content, but it _may_ be surprising behaviour in this case.

wjt (changes requested):
I think this is good, but we would also need to install Kolibri… :)

wjt:
Can we make a separate Maniphest task to track this, please?

config/personality/base.ini
```diff
@@ -12,3 +12,15 @@ apps_del = ${apps_add_defaults}
 [flatpak-remote-flathub]
 # Delete all default additions and explicitly build the list from scratch.
 apps_del = ${apps_add_defaults}
+apps_add =
+  org.learningequality.Kolibri
```
wjt, line 16: This is probably a fine start but I do wonder if we should actually be putting this into `apps_add_mandatory` in `config/defaults.ini`. It is at least as important a part of Endless OS 4 and beyond as Hack is, IMO.

config/personality/base.ini
```diff
@@ -12,3 +12,15 @@ apps_del = ${apps_add_defaults}
...
+
+[kolibri]
+install_channels =
+  # How to get started with Kolibri on Endless OS [23]
+  e8a879742b2249a0a4b890f9903916f7
+
+[kolibri-e8a879742b2249a0a4b890f9903916f7]
+include_node_ids =
+  # English [topic - 23]
+  3b909a18242c48208dbc49d06bc48162
```
wjt, line 26: Perhaps this could live in `config/defaults.ini` too and then would be overridden for `es`, `ar` and `fr` with those versions' node IDs.

dylanmccall: Yeah, we could do that. That exact thing is commented out in config/defaults.ini right now, so we could uncomment that and delete this.

wjt: Let's do that. If you are able to do the `config/personality/{es,ar,fr}.ini` overrides too that would be wonderful :)

dylanmccall: Okay, done :)

dylanmccall:
I'll go ahead and merge this with the bits you noted taken care of :)

PR #19, opened 2021-05-01
Needs rsync

BryanQuigley:
It's part of how mmdebstrap-setup/cleanup helpers work.


dbnicholson (approved):
I completely forgot about that, thanks. I wonder if that should be rewritten to use something more commonly available.

PR #21, opened 2021-05-02
Remove python2 from buildroot

BryanQuigley:
I was able to build images fine without it and don't see any scripts that need it.

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -94,7 +94,6 @@ packages_add =
   p7zip-full
   parted
   pigz
-  python
   python3
   python3-apt
   python3-gi


dbnicholson:
I had a suspicion that python2 was still used in some esoteric setup, and it is in our split image tool. We're actually phasing it out, but in the meantime #22 fixes it. After that's merged I think we can merge this.

jprvita (approved):
LGTM, and after #22 I can't find any other use of python2 in this tool (and if some corner case breaks, we should fix it to use python3 anyway).

jprvita:
Thanks for the contribution @BryanQuigley!

@dbnicholson Feel free to merge this PR when you have a chance.

PR #22, opened 2021-05-04
split-image: Use python3 for disk size calculations

dbnicholson:
This is the last usage of python2 in the image builder. After this it
can be dropped from the buildroot.


jprvita (approved):

jprvita:
Great to get rid of python2 usage in another tool :-)

PR #23, opened 2021-05-15
helpers: Add a way to create a data-only USB image

dylanmccall:
https://phabricator.endlessm.com/T31962

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -294,6 +294,7 @@ include_node_ids =
 [usb]
 size = 16000000000
 free_space = 1000
+live_image = true
 
 [split]
 # Merged customization hooks to run.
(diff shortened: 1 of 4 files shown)


dylanmccall:
This is a completely evil thing to do, and to make matters worse I'm going to merge this at the end of everyone's day on a Friday so we can have a test build. Please feel free to review anyway and I promise I will come back around to clean this up. Very curious about any thoughts for how we can jam this sort of supplemental data-only image into our infrastructure without _too_ much rewriting of things, although I suspect the correct solution may just involve a separate image builder-like thing.

dbnicholson:
@dylanmccall you can run the image builder in Jenkins with an alternate branch if you need to make a year build.

dylanmccall:
Ooh, really! Sorry, I was meaning to ask you about that but ran out of day and I didn't want to bug you about it in your evening. Done, and I finally realized the parameter which does that :)

PR #26, opened 2021-05-19
image: Drop merged flatpak repo setup

dbnicholson:
This was used to ensure that flatpak's repo at `/var/lib/flatpak/repo`
could be merged with `/ostree/repo`. We don't want that configuration
anymore, so it can be dropped.

With the repos disconnected, we no longer have to disable the OS remote
since the flatpak repo would never have it in the first place. Likewise,
no symlink juggling is needed to access the installation outside of the
chroot.

https://phabricator.endlessm.com/T31980


hooks/image/70-flatpak-manifest
```diff
@@ -56,25 +56,10 @@ if os.environ.get('EIB_FLATPAK_ENABLE', 'false') != 'true':
     write_manifest(data, hook_group)
     exit(0)
 
-# Open the flatpak installation in the OS /var
+# Open the flatpak installation in the OS /var/lib/flatpak or split disk
+# /var/endless-extra/flatpak.
```
pwithnall, line 60: Are you planning on dropping the split-disk stuff in a separate MR? aiui we’re dropping that feature too.

dbnicholson: Yeah, separate feature to delete. There's a lot more to the split disk step than just this.

pwithnall: SGTM. Merge when the mood takes you

dbnicholson: The mood has struck me. 👯

PR #30, opened 2021-07-21
config: Do not include org.freedesktop.Platform.openh264 for arm64

jprvita:
This runtime extension is not available for arm64.

https://phabricator.endlessm.com/T30865

### DIFF
--- a/config/arch/arm64.ini
+++ b/config/arch/arm64.ini
@@ -11,6 +11,11 @@ iso = false
 hooks_del = 50-google-chrome
 
 [flatpak-remote-flathub]
+# These extensions are not available for arm64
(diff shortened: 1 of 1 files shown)


config/arch/arm64.ini
```diff
@@ -11,6 +11,10 @@ iso = false
 hooks_del = 50-google-chrome
 
 [flatpak-remote-flathub]
+# These extensions are not available for arm64
```
wjt, line 14:
```suggestion
# These extensions are not available for arm64
# https://gitlab.com/freedesktop-sdk/openh264-extension/-/issues/2
```

wjt: LGTM with that one change.

PR #31, opened 2021-08-16
config: Set eos-apps repo to use eos3 branch by default

jprvita:
The latest eos-apps are available on the eos3 branch of the eos-apps
flatpak repo. We are not develpoing these apps anymore, so it probably
does not make sense to push a eos4 ref to this repo.

https://phabricator.endlessm.com/T32376

### DIFF
--- a/config/branch/master.ini
+++ b/config/branch/master.ini
@@ -3,8 +3,3 @@
 [ostree]
 # Use the dev repo for deployment since master is never released
 deploy_url = ${dev_deploy_repo_url}/${repo}
-
-[flatpak-remote-eos-apps]
-# Endless apps are now published to the eos3 branch rather than master.
-# This default to the series, which would be master here.
-default_branch = eos3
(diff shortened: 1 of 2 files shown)


dbnicholson (changes requested):
I think it would better to change `default_branch` in `defaults.ini` and drop the master override that does the same thing. That way we won't need to carry this forward for every `eosX.Y.ini`. As you say, we're almost certainly never going to develop these again to where we'd use separate branches.

jprvita:
> I think it would better to change `default_branch` in `defaults.ini` and drop the master override that does the same thing. That way we won't need to carry this forward for every `eosX.Y.ini`. As you say, we're almost certainly never going to develop these again to where we'd use separate branches.

Makes sense.

jprvita:
Updated following @dbnicholson's suggestion

dbnicholson (approved):

PR #32, opened 2021-09-10
hooks: Remove 50-google-chrome and default to Chromium instead

pwithnall:
Now we can ship `org.chromium.Chromium.Codecs` out of the box in many
cases, and in cases where we can't, it’ll be downloaded with app updates
relatively quickly when the machine goes online (similar to how Chrome
was silently installed when a network connection became available).

So we can use Chromium as the default browser everywhere, and have a
similar experience.

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

https://phabricator.endlessm.com/T32407


pwithnall:
I left the following two hooks untouched, as we might still want to customise Chrome’s bookmarks and plugins update URI. I can drop them if wanted:
 * `hooks/image/50-chrome-bookmarks-policy`
 * `hooks/image/50-chrome-plugins-updater-url`

It’s possible that we do want to drop them anyway: the bookmarks hook was added in commit c4ccb60397411b2e207b0975a1440ae4a1175959 in 2016 for the Mission in order to preseed bookmarks. The plugins hook was added in 3810acdc365df0672817b4ae0fe36ddcb0b348e7 (and also last touched in 2016) as the normal S3 URI for providing Chrome plugins was not accessible in China.

pwithnall:
I can’t actually find any uses of either of those plugins in endless-image-config, so unless there’s somewhere else they’re used, I think they are dead and should be removed. I can do that in a follow-up PR if wanted.

In particular, the last use of the bookmarks hook was removed in ba56d26913433f50dce3e97d8f81cac78a546ae9, and the last use of the plugins hook was removed in 5da2f2cac0586855cefe26ba398592d02b9c8457 (both in endless-image-config.git).

wjt:
Chrome no longer has plugins – RIP Flash – so we can certainly delete that one. I suggest removing both.

PR #34, opened 2021-09-20
config: add Kolibri content to the en personality

dylanmccall:
https://phabricator.endlessm.com/T32522


wjt (approved):
Sure, I think this makes sense as a bare minimum to make en a superset of base. Is this content available in any other languages?

dylanmccall:
> Sure, I think this makes sense as a bare minimum to make en a superset of base. Is this content available in any other languages?

Yes, it is. This is the earlier commit, which added the content to base.ini, and specified different selections for es.ini and fr.ini…

d8ed0ed065c9b8df0840b8f70624739c3139520e

… and I just realized an error, there: we don't set `install_channels` for those languages either :/

PR #36, opened 2021-09-29
config: add Kolibri content to the en personality

dylanmccall:
https://phabricator.endlessm.com/T32522


wjt:
If the build test on master succeeds please fast-forward eos4.0 to match master.

PR #39, opened 2021-10-07
config: Use Kolibri beta for master branch builds

dylanmccall:
https://phabricator.endlessm.com/T32629

### DIFF
--- a/config/branch/master.ini
+++ b/config/branch/master.ini
@@ -6,3 +6,14 @@ deploy_url = ${dev_deploy_repo_url}/${repo}
 
 # Master images should always follow the master ostree.
 stable_branch = ${build:branch}
+
+[flatpak-remote-flathub]
+# Don't install the stable version of Kolibri
+apps_del =
+  org.learningequality.Kolibri
+
+[flatpak-remote-flathub-beta]
+# Instead, install the beta version of Kolibri
+enable = true
+apps_add =
+  org.learningequality.Kolibri
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -487,6 +487,11 @@ apps_add =
   org.pitivi.Pitivi
   org.tuxpaint.Tuxpaint
 
+[flatpak-remote-flathub-beta]
+# Flathub Beta repo, disabled by default
+repo_file = https://dl.flathub.org/beta-repo/flathub-beta.flatpakrepo
+enable = false
+
 # Convenience variable primarily for other configurations to add or
 # delete runtimes from the default set
 runtimes_add =
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -416,6 +416,9 @@ enable_p2p_updates = true
 # default_branch (optional) - The default branch to use for installs.
 #   This affects installs during the image build as well as at runtime.
 #   Flatpaks that do not specify a branch will use this by default.
+# prio (optional) - The priority for the remote. (See xa.prio).
+# noenumerate (optional) - Whether this remote should be used when searching
+#   for available apps or runtimes. (See xa.noenumerate).
 #
 # Additionally, a URL or path for a flatpakrepo file can be specified in
 # the repo_file option. The Url, Title, DefaultBranch and GPGKey keys
@@ -487,11 +490,6 @@ apps_add =
   org.pitivi.Pitivi
   org.tuxpaint.Tuxpaint
 
-[flatpak-remote-flathub-beta]
-# Flathub Beta repo, disabled by default
-repo_file = https://dl.flathub.org/beta-repo/flathub-beta.flatpakrepo
-enable = false
-
 # Convenience variable primarily for other configurations to add or
 # delete runtimes from the default set
 runtimes_add =
@@ -533,3 +531,11 @@ allow_extra_data_add =
   #
   # https://phabricator.endlessm.com/T27888
   org.freedesktop.Platform.openh264
+
+[flatpak-remote-flathub-beta]
+# Flathub Beta repo, disabled by default
+repo_file = https://dl.flathub.org/beta-repo/flathub-beta.flatpakrepo
+enable = false
+default_branch = beta
+prio = 0
+noenumerate = true
--- a/lib/eibflatpak.py
+++ b/lib/eibflatpak.py
@@ -22,6 +22,7 @@ import base64
 import codecs
 from collections import namedtuple, OrderedDict
 from configparser import ConfigParser
+from contextlib import contextmanager
 import eib
 from eibostree import fetch_remote_collection_id
 import fnmatch
@@ -149,7 +150,7 @@ class FlatpakRemote(object):
     def __init__(self, manager, name, url=None, deploy_url=None,
                  repo_file=None, apps=None, runtimes=None, exclude=None,
                  allow_extra_data=None, title=None, default_branch=None,
-                 **extra_options):
+                 prio=None, noenumerate=None, **extra_options):
         # Copy some manager attributes
         self.manager = manager
         self.installation = manager.installation
@@ -167,6 +168,8 @@ class FlatpakRemote(object):
             if allow_extra_data else set()
         self.title = title
         self.default_branch = default_branch
+        self.prio = prio
+        self.noenumerate = noenumerate
 
         # Only supported from repo_file
         self.gpg_key = None
@@ -233,6 +236,8 @@ class FlatpakRemote(object):
             remote.set_title(self.title)
         if self.default_branch:
             remote.set_default_branch(self.default_branch)
+        if self.prio is not None:
+            remote.set_prio(self.prio)
 
         # Import the GPG key if specified
         if self.gpg_key:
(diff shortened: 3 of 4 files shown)


config/branch/master.ini
```diff
@@ -6,3 +6,14 @@ deploy_url = ${dev_deploy_repo_url}/${repo}
...
 stable_branch = ${build:branch}
+
+[flatpak-remote-flathub]
+# Don't install the stable version of Kolibri
+apps_del =
+  org.learningequality.Kolibri
+
+[flatpak-remote-flathub-beta]
+# Instead, install the beta version of Kolibri
+repo_file = https://dl.flathub.org/beta-repo/flathub-beta.flatpakrepo
```
dbnicholson, line 17: Even though master is probably the only place we'd ever use the flathub beta repo, I'd prefer if it was defined in `defaults.ini` with `enable = false`. Like:
```
[flatpak-remote-flathub-beta]
repo_file = https://dl.flathub.org/beta-repo/flathub-beta.flatpakrepo
# Disabled by default
enable = false
```
Then in `master.ini` just add `enable = true`.

But I would actually split that into a separate commit since enabling the beta repo now means it's available all the time. Ideally we would set `xa.prio` to 0 (or maybe even `xa.noenumerate` to `true`, but I'm not totally sure what the affects) so that you still use the normal flathub repo by default. But that would require some more plumbing. Maybe it's OK if master images are setup to prefer flathub beta releases at runtime.

dylanmccall: Oh yeah, that's tidier. Thanks! :)

config/branch/master.ini
```diff
@@ -6,3 +6,14 @@ deploy_url = ${dev_deploy_repo_url}/${repo}
...
+[flatpak-remote-flathub]
+# Don't install the stable version of Kolibri
+apps_del =
+  org.learningequality.Kolibri
+
+[flatpak-remote-flathub-beta]
+# Instead, install the beta version of Kolibri
+enable = true
+apps_add =
+  org.learningequality.Kolibri//beta
```
wjt, line 19: Is the `//beta` suffix needed?

I see that we define a `default_branch` for regular Flathub and eos-apps. Should we be setting that in the `defaults.ini` definition of flathub-beta?

dylanmccall:
Okay, I changed this to set `xa.prio=0` and `xa.noenumerate=true` for the flathub-beta remote. This way, _only_ the apps we select will be installed from there, and flathub-beta won't appear as an option to install things from in the final image. To achieve this I needed to add a bit of extra code to `eibflatpak.py` since we need to at least allow it to enumerate the flatpak remote while we're building the image.

I'm pretty happy with it now :)

config/defaults.ini
```diff
@@ -416,6 +416,7 @@ enable_p2p_updates = true
 # default_branch (optional) - The default branch to use for installs.
 #   This affects installs during the image build as well as at runtime.
 #   Flatpaks that do not specify a branch will use this by default.
+# prio (optional) - The priority for the remote (see xa.prio).
```
dbnicholson, line 419: Could you add a comment for `noenumerate` here, too? Might also be nice to add a reference to flatpak-remote(5), which you can view online at https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpak-remote.

lib/eibflatpak.py
```diff
@@ -344,17 +351,36 @@ def reset_metadata(self):
...
+            yield
+        else:
+            remote = self.installation.get_remote_by_name(self.name)
+            try:
+                remote.set_noenumerate(False)
+                self.installation.modify_remote(remote)
+                yield
+            finally:
+                remote.set_noenumerate(self.noenumerate)
+                self.installation.modify_remote(remote)
```
dbnicholson, line 372: Is the problem here that calling `list_remote_refs_sync` on a remote with `noenumerate = true` fails? Another option is to treat `noenumerate` specially and handle it in the `deploy` method where it would only get set at the end.

lib/eibflatpak.py
```diff
@@ -344,17 +351,36 @@ def reset_metadata(self):
             remote.set_title(self.title)
         if self.default_branch:
             remote.set_default_branch(self.default_branch)
+        if self.prio is not None:
+            remote.set_prio(self.prio)
+        if self.noenumerate is not None:
+            remote.set_noenumerate(self.noenumerate)
```
dbnicholson, line 357: I think you'd also need to include `self.prio` and `self.noenumerate` into the early return at the beginning of this method.

dbnicholson (commented):
Thanks for digging into the mess that is `eibflatpak`! I had a couple questions but I think this is a really useful addition.

config/defaults.ini
dylanmccall: Okay, done!

lib/eibflatpak.py
dylanmccall: Oh, I like that idea! I ended up with this diff, although I need to test it on Monday before I push a commit somewhere. I don't think I properly understand what `reset_metadata` is for, so I might have broken something here :)

```diff
diff --git a/lib/eibflatpak.py b/lib/eibflatpak.py
index 7aed1ad..4870be4 100644
--- a/lib/eibflatpak.py
+++ b/lib/eibflatpak.py
@@ -238,8 +238,6 @@ class FlatpakRemote(object):
             remote.set_default_branch(self.default_branch)
         if self.prio is not None:
             remote.set_prio(self.prio)
-        if self.noenumerate is not None:
-            remote.set_noenumerate(self.noenumerate)
 
         # Import the GPG key if specified
         if self.gpg_key:
@@ -303,12 +301,7 @@ class FlatpakRemote(object):
         Sync the appstream and metadata from the remote. If the deploy
         URL differs from the pull URL, it's adjusted here, too.
         """
-        if self.deploy_url != self.url:
-            logger.info('Setting %s URL to %s for deployment', self.name,
-                        self.deploy_url)
-            remote = self.installation.get_remote_by_name(self.name)
-            remote.set_url(self.deploy_url)
-            self.installation.modify_remote(remote)
+        self._modify_remote_for_deployment()
 
         # Set the flatpak remote collection ID if it's enabled and the
         # remote has a collection ID
@@ -335,15 +328,37 @@ class FlatpakRemote(object):
         # Reset any configuration defined metadata
         self.reset_metadata()
 
+    def _modify_remote_for_deployment(self):
+        if self.deploy_url == self.url and self.noenumerate is None:
+            return
+
+        remote = self.installation.get_remote_by_name(self.name)
+
+        if self.deploy_url != self.url:
+            logger.info('Setting %s URL to %s for deployment', self.name,
+                        self.deploy_url)
+            remote.set_url(self.deploy_url)
+        
+        if self.noenumerate is not None:
+            logger.info('Setting %s noenumerate to %s for deployment',
+                        self.name, self.noenumerate)
+            remote.set_noenumerate(self.noenumerate)
+
+        self.installation.modify_remote(remote)
+
     def reset_metadata(self):
         """Reset remote's metadata per configuration
 
-        Set the title and default branch remote options back to the
+        Set, title, default branch, and prio remote options back to the
         value used in the builder configuration. This is needed if the
-        setting was changed by a call to update the remote's metadata
-        from the server.
+        setting was changed by a call to update the remote's metadata from
+        the server.
         """
-        if not self.title and not self.default_branch:
+        if (
+            not self.title
+            and not self.default_branch
+            and self.prio is None
+        ):
             return
 
         remote = self.installation.get_remote_by_name(self.name)
@@ -353,24 +368,8 @@ class FlatpakRemote(object):
             remote.set_default_branch(self.default_branch)
         if self.prio is not None:
             remote.set_prio(self.prio)
-        if self.noenumerate is not None:
-            remote.set_noenumerate(self.noenumerate)
         self.installation.modify_remote(remote)
 
-    @contextmanager
-    def _allow_enumerate(self):
-        if not self.noenumerate:
-            yield
-        else:
-            remote = self.installation.get_remote_by_name(self.name)
-            try:
-                remote.set_noenumerate(False)
-                self.installation.modify_remote(remote)
-                yield
-            finally:
-                remote.set_noenumerate(self.noenumerate)
-                self.installation.modify_remote(remote)
-
     def enumerate(self):
         """Populate refs from remote data
 
@@ -378,9 +377,8 @@ class FlatpakRemote(object):
         app or runtime found.
         """
         logger.info('Fetching refs for %s', self.name)
-        with self._allow_enumerate():
-            all_remote_refs = eib.retry(
-                self.installation.list_remote_refs_sync, self.name)
+        all_remote_refs = eib.retry(
+            self.installation.list_remote_refs_sync, self.name)
         for remote_ref in all_remote_refs:
             # Get the full ostree ref
             ref = remote_ref.format_ref()
```

dylanmccall:
Okay, I changed this to special-case noenumerate in the eibflatpak's `deploy` method, as @dbnicholson suggested, and that does indeed work quite nicely :)

dbnicholson:
Sorry, been busy the past week but I'll review in the next day or so.

dbnicholson (approved):
Finally got around to this. It looks good to me and I tested it out locally. Seems to work correctly. I wonder if just setting a lower priority without noenumerate is good enough, but I think we can merge this and see how it goes.

PR #42, opened 2021-11-04
config: Adjust pre-loaded Kolibri content on base, en, es, fr and pt_BR images

jprvita:
Drops outdated "Kolibri on Endless OS" documentation channel and includes localized Khan Academy content and the TED-Ed Kolibri channel.

https://phabricator.endlessm.com/T32693

### DIFF
--- a/config/personality/es.ini
+++ b/config/personality/es.ini
@@ -40,6 +40,11 @@ apps_add =
   com.endlessm.water_and_sanitation.es
   com.endlessm.world_literature.es
 
+[kolibri]
+install_channels =
+  # Khan Academy (Español)
+  c1f2b7e6ac9f56a2bb44fa7a48b66dce
+
 [kolibri-e8a879742b2249a0a4b890f9903916f7]
 include_node_ids =
   # Español [topic]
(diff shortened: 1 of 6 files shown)


config/personality/es.ini
```diff
@@ -40,7 +40,14 @@ apps_add =
...
+[kolibri]
+install_channels =
+  # Khan Academy (Español)
+  c1f2b7e6ac9f56a2bb44fa7a48b66dce
+
 [kolibri-e8a879742b2249a0a4b890f9903916f7]
 include_node_ids =
   # Español [topic]
   6e8f60c6b9c841969853d48f4eff22cf
+  # TED-Ed Endless Key [complete]
```
wjt, line 52: Does this channel include Spanish (respectively French below) content? I don't think it makes sense to include it in these lang-specific images otherwise.

I think this is wrong anyway because you are specifying the channel ID of TED-Ed, but as a node ID of the "How to get started" channel which we are not including.

wjt (changes requested):

config/personality/es.ini
jprvita: It does not, only English. However, I believe the content is valuable anyway, as (at least in my experience growing up in Brazil) there is really no way around English content if you are trying to get into programming. So I think we should include it, but I'll leave it out of this PR and we can discuss that separately.

PR #43, opened 2021-11-04
pt_BR: Drop inexistent app com.endlessm.extra.pt_BR

jprvita:
This app does not exist on eos-apps.

https://phabricator.endlessm.com/T32693


dbnicholson (approved):
Did it at one point? Who knows...

PR #46, opened 2021-11-06
kolibri: Clear local dir before downloading content

jprvita:
When the hook that downloads Kolibri content runs, its target directory
may already exist and have content from the previous run, which may be
from a different image configuration.

This commit removes any previously exiting content to prevent that
content from being included in the image being built.

https://phabricator.endlessm.com/T32712

### DIFF
--- a/hooks/image/60-kolibri-content
+++ b/hooks/image/60-kolibri-content
@@ -27,6 +27,12 @@ import_kolibri_channel()
 
 # Needs to be kept in sync with hooks/image/61-kolibri-content-install
 KOLIBRI_CONTENT_DIR="${EIB_CONTENTDIR}/kolibri-content"
+# FIXME: For now, we need to remove old content that may exist from previous
+# runs to prevent images accidentally getting extra channels that are not in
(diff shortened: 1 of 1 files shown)


dbnicholson:
Sorta defeats the purpose of the entire thing. If the caching can't be made to work correctly then maybe we should just combine the install hook with this hook? I think you'd just change `KOLIBRI_HOME` so it's set to `"${OSTREE_VAR}"/lib/kolibri/data` and drop all the `KOLIBRI_CONTENT_DIR` handling. I'd like to hear what @dylanmccall has to say, though, since he wrote this. I've never actually tried to provision kolibri content.

dbnicholson:
Can you update the comment and commit message to reflect that? The way it sounds from the commit is that this is just something you want to do and doesn't express what the implications are.

jprvita:
> Can you update the comment and commit message to reflect that? The way it sounds from the commit is that this is just something you want to do and doesn't express what the implications are.

Makes sense, I've done that now.

dbnicholson (approved):

PR #47, opened 2021-12-13
kolibri: Add options for desktop auth plugin

manuq:
Add option for partners that wish to force Kolibri superadmin user for
regular system users.

If the option is set in the image INI file, it will be added to
Kolibri's options INI file.

https://phabricator.endlessm.com/T33035

### DIFF
--- a/hooks/image/60-kolibri-content
+++ b/hooks/image/60-kolibri-content
@@ -41,6 +41,7 @@ source ${venv_dir}/bin/activate
 
 pip install kolibri==${EIB_KOLIBRI_APP_VERSION}
 pip install kolibri-app-desktop-xdg-plugin==${EIB_KOLIBRI_APP_DESKTOP_XDG_PLUGIN_VERSION}
+pip install kolibri-desktop-auth-plugin==${EIB_KOLIBRI_DESKTOP_AUTH_PLUGIN_VERSION}
 
 export KOLIBRI_HOME="${KOLIBRI_CONTENT_DIR}"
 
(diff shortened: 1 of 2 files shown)


manuq:
Marking as draft because this depends on https://github.com/endlessm/kolibri-desktop-auth-plugin/pull/9 and on releasing that plugin to PyPi.

dylanmccall:
It kind of looks like this needs a new Phabricator ticket, but we can work on merging this now.

danigm (approved):

hooks/image/60-kolibri-content
```diff
@@ -50,9 +51,16 @@ cat << EOF > "${KOLIBRI_HOME}"/options.ini
 [Deployment]
 STATIC_USE_SYMLINKS = 0
 EOF
+if [ "${EIB_KOLIBRI_DESKTOP_AUTH_PLUGIN_REGULAR_USERS_CAN_MANAGE_CONTENT}" = "true" ]; then
+    cat << EOF >> "${KOLIBRI_HOME}"/options.ini
+[DesktopAuth]
+REGULAR_USERS_CAN_MANAGE_CONTENT = True
+EOF
+fi
```
dylanmccall, line 59: This hook is only going to run if `EIB_KOLIBRI_INSTALL_CHANNELS` is set, which is not the case for all of the images where we expect `EIB_KOLIBRI_DESKTOP_AUTH_PLUGIN_REGULAR_USERS_CAN_MANAGE_CONTENT` to be true. Instead, I think we should do this in its own hook, similar to #55.

hooks/image/60-kolibri-content
```diff
@@ -50,9 +51,16 @@ cat << EOF > "${KOLIBRI_HOME}"/options.ini
...
 
 kolibri plugin enable kolibri.plugins.app
 kolibri plugin enable kolibri_app_desktop_xdg_plugin
+kolibri plugin enable kolibri_desktop_auth_plugin
```
dylanmccall, line 63: :+1:

dylanmccall (changes requested):

dylanmccall:
I built on top of this in #60, changing the option name and using a new hook to generate our options.ini file.

PR #49, opened 2021-12-21
Kolibri automatic provisioning: Change default settings

manuq:
Add more facility and device settings. Use defaults more proper for
the seamless user management case.

https://phabricator.endlessm.com/T32885

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -244,6 +244,11 @@ automatic_provision_superuser_name =
 automatic_provision_superuser_password =
 automatic_provision_preset = nonformal
 automatic_provision_landing_page = learn
+automatic_provision_learner_can_edit_username = false
+automatic_provision_learner_can_edit_name = false
+automatic_provision_learner_can_edit_password = false
+automatic_provision_learner_can_sign_up = false
+automatic_provision_allow_guest_access = false
 
 # Preloaded Kolibri channels
 #
--- a/hooks/image/61-kolibri-content-install
+++ b/hooks/image/61-kolibri-content-install
@@ -17,9 +17,15 @@ if [ "${EIB_KOLIBRI_AUTOMATIC_PROVISION}" = "true" ]; then
   "superusername": "${EIB_KOLIBRI_AUTOMATIC_PROVISION_SUPERUSER_NAME}",
   "superuserpassword": "${EIB_KOLIBRI_AUTOMATIC_PROVISION_SUPERUSER_PASSWORD}",
   "preset": "${EIB_KOLIBRI_AUTOMATIC_PROVISION_PRESET}",
-  "facility_settings": {},
+  "facility_settings": {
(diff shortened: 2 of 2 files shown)


config/defaults.ini
```diff
@@ -244,6 +244,11 @@ automatic_provision_superuser_name =
 automatic_provision_superuser_password =
 automatic_provision_preset = nonformal
 automatic_provision_landing_page = learn
+automatic_provision_learner_can_edit_username = false
+automatic_provision_learner_can_edit_name = false
+automatic_provision_learner_can_edit_password = false
+automatic_provision_learner_can_sign_up = false
+automatic_provision_allow_guest_access = false
```
dylanmccall, line 251: I'll be happy with this if we can keep these settings as they are upstream (_unset_ if it is possible?). Instead, something like `learner_can_edit_name = false` should be in the configuration for a specific image.

hooks/image/61-kolibri-content-install
```diff
@@ -17,9 +17,15 @@ if [ "${EIB_KOLIBRI_AUTOMATIC_PROVISION}" = "true" ]; then
...
-  "facility_settings": {},
+  "facility_settings": {
+    "learner_can_edit_username": ${EIB_KOLIBRI_AUTOMATIC_PROVISION_LEARNER_CAN_EDIT_USERNAME},
+    "learner_can_edit_name": ${EIB_KOLIBRI_AUTOMATIC_PROVISION_LEARNER_CAN_EDIT_NAME},
+    "learner_can_edit_password": ${EIB_KOLIBRI_AUTOMATIC_PROVISION_LEARNER_CAN_EDIT_PASSWORD},
+    "learner_can_sign_up": ${EIB_KOLIBRI_AUTOMATIC_PROVISION_LEARNER_CAN_SIGN_UP}
+  },
   "device_settings": {
     "landing_page": "${EIB_KOLIBRI_AUTOMATIC_PROVISION_LANDING_PAGE}",
+    "allow_guest_access": ${EIB_KOLIBRI_AUTOMATIC_PROVISION_ALLOW_GUEST_ACCESS},
```
dylanmccall, line 28: :+1: I'm glad this works!

dylanmccall (changes requested):

config/defaults.ini
manuq: Yeah good point. It's a bit of a pain to build a JSON from Bash, but you are right.

manuq: After a second though I tend to disagree, see my comment in phabricator. https://phabricator.endlessm.com/T32885#919180

manuq: One thing we could do is decuple the device "can guests access" setting from the rest which are valid in all cases.

dylanmccall (approved):
Sorry, this all got buried in Phabricator. I think I'm happy with this. A couple of the defaults still feel a bit funky to me, but they're just that: defaults. And we have a matching Kolibri release now so it should all work nicely. Let's merge it and see how it goes :)

PR #55, opened 2022-01-28
T32885 split automatic provision

dylanmccall:
https://phabricator.endlessm.com/T32885


dbnicholson (approved):
Thanks @dylanmccall! I'm pretty ignorant about Kolibri but it does indeed look better to separate that process.

PR #62, opened 2022-03-24
Hooks: Update format for Kolibri automatic provisioning

dylanmccall:
https://phabricator.endlessm.com/T33314


dylanmccall:
I'm marking this pull request as a draft because we can't merge it until the Kolibri flatpak with Kolibri 0.15 is released.

wjt:
What will happen if an image built with the old configuration gets an upgrade to the new kolibri?

manuq (approved):
+1 the other fields don't need changes.

Issue #63, opened 2022-03-31
Fails if `/var/cache/apt/archives/` is missing


cassidyjames:
Running on a fresh install of Endless OS 4.0.4, `eos-image-builder` fails because `/var/cache/apt/archives/` is missing.

```
$ sudo ./eos-image-builder 
[sudo] password for cassidyjames: 
+ 15:16:11 INFO run-build: Configuring image build
+ 15:16:11 INFO run-build: Loaded configuration file /sysroot/home/cassidyjames/Projects/endlessm/eos-image-builder/config/defaults.ini
+ 15:16:11 INFO run-build: Loaded configuration file /sysroot/home/cassidyjames/Projects/endlessm/eos-image-builder/config/product/eoscustom.ini
+ 15:16:11 INFO run-build: Loaded configuration file /sysroot/home/cassidyjames/Projects/endlessm/eos-image-builder/config/branch/master.ini
+ 15:16:11 INFO run-build: Loaded configuration file /sysroot/home/cassidyjames/Projects/endlessm/eos-image-builder/config/arch/amd64.ini
+ 15:16:11 INFO run-build: Loaded configuration file /sysroot/home/cassidyjames/Projects/endlessm/eos-image-builder/config/personality/base.ini
+ 15:16:11 INFO run-build: Loaded configuration file /sysroot/home/cassidyjames/Projects/endlessm/eos-image-builder/config/branch-personality/master-base.ini
+ 15:16:11 INFO run-build: Preparing build directory
+ 15:16:11 INFO run-build: Cleaning up /var/tmp/eos-image-builder
+ 15:16:11 INFO eib: Killing processes in /var/tmp/eos-image-builder
+ 15:16:11 INFO eib: Unmounting filesystems in /var/tmp/eos-image-builder
+ 15:16:11 INFO run-build: Cleaning up /var/cache/eos-image-builder
+ 15:16:11 INFO eib: Killing processes in /var/cache/eos-image-builder
+ 15:16:11 INFO eib: Unmounting filesystems in /var/cache/eos-image-builder
+ 15:16:11 INFO run-build: Ensuring all ostree checkouts mutable in /var/cache/eos-image-builder/tmp/ostree-co
+ 15:16:11 INFO run-build: Recreating build specific temporary directories
+ 15:16:11 INFO run-build: Exporting configuration to environment
+ 15:16:11 INFO run-build: Output merged configuration file to /var/cache/eos-image-builder/tmp/config.ini
[...]

PR #64, opened 2022-03-31
ImageBuildRoot: Disable buildroot apt cache handling

dbnicholson:
The mmdebstrap hooks were doing entirely the wrong thing by syncing
between the image builder's apt cache and the **system's**
`/var/cache/apt/archives`. That's both wrong and pointless.

Changing that to sync into the buildroot's apt archives makes
`mmdebstrap` throw an error. Apparently it uses the contents of
`/var/cache/apt/archives` to figure out what to install after telling
apt to download packages, so it requires the directory to be empty when
it starts.

To bring this back, something like apt-cacher that works at the network
level is needed.

Fixes: #63

This should be backported to the eos4.0 branch.

### DIFF
--- a/run-build
+++ b/run-build
@@ -88,13 +88,19 @@ class ImageBuildRoot(object):
         keyring = eib.get_keyring(config)
 
         # Wipe the apt cache directory if it's grown too large.
-        if os.path.exists(aptcache_dir):
-            aptcache_size = eib.disk_usage(aptcache_dir)
-            if aptcache_size >= aptcache_max_size:
-                log.info('Apt cache directory uses %d bytes, removing',
-                         aptcache_size)
-                shutil.rmtree(aptcache_dir)
-        os.makedirs(os.path.join(aptcache_dir), exist_ok=True)
+        #
+        # FIXME: This is currently commented because mmdebstrap throws
+        # an error if you seed /var/cache/apt/archives in the root
+        # before it runs. To make this work something like apt-cacher is
+        # needed to handle it at the HTTP level.
+        #
+        # if os.path.exists(aptcache_dir):
+        #     aptcache_size = eib.disk_usage(aptcache_dir)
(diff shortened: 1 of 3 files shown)


run-build
```diff
@@ -88,6 +88,11 @@ class ImageBuildRoot(object):
         keyring = eib.get_keyring(config)
 
         # Wipe the apt cache directory if it's grown too large.
+        #
+        # FIXME: This is currently doing nothing because mmdebstrap
+        # throws an error if you seed /var/cache/apt/archives in the
+        # root before it runs. To make this work something like
+        # apt-cacher is needed to handle it it the network level.
```
jprvita, line 95: Shouldn't we remove this block of dead code then? IIUC this will never be pre-seeded if we cache in the network level, apt will still download the packages even if from a server on the same host or local network.

jprvita (approved):
Left one small suggestion. Other than that, LGTM!

run-build
dbnicholson: I think cache at the network level was poorly worded. It would be handled by proxying the HTTP requests and returning `.deb` files from disk or saving them when they're new. Basically a little local CDN.

That said, I don't think I'll be working on it any time soon. Maybe comment out the code for now?

jprvita: Yes, that is what I had in mind. I guess the question is, would we still want to go and clean the proxy cache directory in that case right from eos-image-builder? I thought most proxies would control that themselves.

dbnicholson: I suppose they could, but I think eos-image-builder would also be the one starting and stopping the proxy. In that case, the brute force cleanup would still be reasonable to have.

However, since I don't really know what it would look like, I'll just comment it out for now.

dbnicholson:
Also cherry-picked to eos4.0 in 12bd03032d1edf968f708a166cd5c7f71cd07b5a.

PR #65, opened 2022-04-04
kolibri: Use our content CDN by default

dbnicholson:
To offload our Learning Equality's CDN, we've setup our own CDN in front
of it at https://kolibri-content.endlessos.org. Import channels using
that as the base URL and configure deployments to use it. The
configuration can be cleared to go back to the default URL.

https://phabricator.endlessm.com/T30950


dbnicholson:
This is untested. I decided not to go all the way and configure the deployed options to use our CDN. I have a commit locally that also updates `options.ini`.

danigm (approved):
LGTM

PR #67, opened 2022-04-14
Update Kolibri module versions

dylanmccall:
The newest version of the Kolibri flatpak uses these versions.

https://phabricator.endlessm.com/T33318

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -237,9 +237,9 @@ branding_fbe_config = ${build:datadir}/branding/gnome-initial-setup/default/gnom
 branding_subst_vars_add =
 
 [kolibri]
-app_version = 0.14.7
-app_desktop_xdg_plugin_version = 1.0.9
-desktop_auth_plugin_version = 0.0.6
+app_version = 0.15.2
+app_desktop_xdg_plugin_version = 1.1.4
+desktop_auth_plugin_version = 0.0.7
 
 automatic_provision = false
 automatic_provision_facility_name = Endless


dylanmccall:
We will also need to cherry-pick this to eos4.0.

danigm (approved):
LGTM

dbnicholson:
Questions:

* Do we still want this?
* What's the relation between these versions and the ones in the flatpak?
* Can we bump Kolibri to 0.15.2 so it includes the `importcontent` options I added upstream?

dylanmccall:
[Reiterating from Slack so it doesn't get lost to time]

Yes, we still need this for the time being. The version of Kolibri and any third party plugins doesn't _need_ to match the flatpak app exactly, but it is better if it does. This way when Kolibri starts for the first time on the target system, it can just work straight away. Otherwise it needs to run migrations on the pre-installed database, which can be slow. (And if we run a newer version of Kolibri in the image builder than the desktop app, it's an ugly situation, so we definitely need to avoid that).

It would be nice if we could get eos-image-builder using the Kolibri flatpak directly instead of the Python package. I believe we actually did have it doing that at some point but there was some troublesome stuff with the chroot that caused me to give up on it and fetch the thing from pypi. But it's probably worth revisiting.

I updated this pull request to use Kolibri 0.15.2, which the app will be using after we've released this update on Flathub: https://github.com/flathub/org.learningequality.Kolibri/pull/44.

dbnicholson:
I ran this locally for some other work I was doing that depends on and it worked fine. I'm going to merge it.

wjt:
> We will also need to cherry-pick this to eos4.0.

Where will we track this to ensure it is not forgotten?

PR #68, opened 2022-04-28
New 'micro' ultra-minimal personality

wjt:
The 'base' image has grown and grown. Here's a new 'micro' personality, inspired by @cassidyjames's recent work on a slimmed-down image that contains only a browser, which doesn't even contain a browser. My intention is only to build this on master.

### DIFF
--- a/config/branch-personality/master-micro.ini
+++ b/config/branch-personality/master-micro.ini
@@ -1,0 +1,4 @@
+# Settings for micro personality with master branch
+[flatpak-remote-flathub-beta]
+# No Kolibri plz
+apps =
--- a/config/personality/micro.ini
+++ b/config/personality/micro.ini
@@ -1,0 +1,22 @@
+# Bare-minimum personality with no Flatpaks installed, not even a browser
+
+[image]
+iso = true
+
+hooks_add =
+  50-autoinstall-mask-rhythmbox
+  50-autoinstall-mask-cheese
+
+[flatpak-remote-eos-sdk]
+# Delete the otherwise-mandatory EknServicesMultiplexer that helps with Shell
+# search; this is fine as there isn't space for any searchable content, anyway
+apps =
+
+[flatpak-remote-eos-apps]
+# Completely override the set of installed eos-apps apps
+apps =
+
+[flatpak-remote-flathub]
+# Completely override the set of installed Flathub apps and runtimes
+apps =
+runtimes =
--- a/data/branding/gnome-initial-setup/default/gnome-initial-setup.conf
+++ b/data/branding/gnome-initial-setup/default/gnome-initial-setup.conf
@@ -1,5 +1,2 @@
-[pages]
-run_welcome_tour=false
-
 [Language]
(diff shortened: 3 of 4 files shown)


wjt:
https://ci.endlessos.org/job/image-build-amd64/23083/console nope, need to avoid even more mandatory flatpaks being complained about

wjt:
I had thought that eos-updater-flatpak-installer had a mode where it would write stamp files without actually performing the stated actions, but apparently not. A shame, that would be nice.

jprvita:
Should we instead use a more descriptive name like `noapps`, `systemonly`, `nocontent`, `bare` etc ?

(listening to https://www.youtube.com/watch?v=KwvWtZl2ICY as I write this -- :bike::bike::bike::bike::bike::bike: )

jprvita:
> Should we instead use a more descriptive name like `noapps`, `systemonly`, `nocontent`, `bare` etc ?
I vote for `bare` or `noapps`.

dbnicholson:
I personally like micro even though it's implemented as no apps. But to me it's whatever gets you the smallest image possible that's still Endless OS.

dbnicholson:
Just to throw out one other possibility, in container world Debian provides `slim` variants. I personally like `micro` better.

wjt:
nano? pico? femto? planck

jprvita:
I think we should actually make `base` just be the no-apps-no-content image and give the "small image with a sample of apps and content" something else? What I don't like about the micro / mini / nano / slim is that it has no reference to what is included, and all names just basically mean "a little bit" in the same way, in this context. OTOH, `base` or `bare` to me indicates that this is some fundamental part that can be built upon, and in our case is what we use when we build other images. I think `base` is a great name for that.

cassidyjames:
I think I agree with @jprvita; the naming of smaller-override type images feels like `design_final_2-for-real_USE-THIS-ONE.psd` whereas it would make sense 

:bike::derelict_house: contribution, `core` could be a decent name for the just-the-os-bits image, but I'm not too picky.

wjt:
> I think we should actually make `base` just be the no-apps-no-content image and give the "small image with a sample of apps and content" something else? What I don't like about the micro / mini / nano / slim is that it has no reference to what is included, and all names just basically mean "a little bit" in the same way, in this context. OTOH, `base` or `bare` to me indicates that this is some fundamental part that can be built upon, and in our case is what we use when we build other images. I think `base` is a great name for that.

I agree in principle.

The 'base' personality gets special treatment in the endless installer for windows, which displays side-by-side the choice of "basic" and full (with a dropdown for language). Base is chosen more than half the time.

Capture d’écran de 2022-05-03 09-23-27

It's also the top option on our website for all other installation methods. An image that doesn't include a web browser, or EknServicesMultiplexer (without which desktop search for endless SDK apps won't work even after you install such an an app), is an even worse demo of what Endless OS is about than the base image is today!

So if we want to use the 'base' name to mean 'the OSTree and nothing else' we need to:

- Create a replacement small-but-somewhat-representative image
- Release it
- Adjust the website, wiki, endless installer for windows, release scripts, etc. to use it instead of base
- And only now can we repurpose 'base'

Not saying we shouldn't do that; just explaining why I did not want to do that for this quick proof-of-concept which doesn't even work :)

PR #69, opened 2022-04-29
Remove armhf support

wjt:
eos3.8 was the last branch to support this architecture & collection of
platforms.

https://phabricator.endlessm.com/T30297


dbnicholson (approved):
I didn't look closely myself, but that looks accurate.

PR #71, opened 2022-05-10
Kolibri importcontent fixes

dbnicholson:
The point of this is to make installing channel contents reliable. However, this is a draft for now since it depends on the not-yet-released 0.15.3.

https://phabricator.endlessm.com/T33363
https://phabricator.endlessm.com/T33365

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -237,7 +237,7 @@ branding_fbe_config = ${build:datadir}/branding/gnome-initial-setup/default/gnom
 branding_subst_vars_add =
 
 [kolibri]
-app_version = 0.15.2
+app_version = 0.15.3
 app_desktop_xdg_plugin_version = 1.1.4
 desktop_auth_plugin_version = 0.0.7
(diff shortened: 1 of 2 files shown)


dbnicholson:
0.15.3 got released, so this is ready for review now.

dylanmccall:
This looks good to me, but we'll need to update the version of Kolibri in https://github.com/flathub/org.learningequality.Kolibri, first. Otherwise the newer database will cause trouble with the old Kolibri in our target system.

dbnicholson:
> This looks good to me, but we'll need to update the version of Kolibri in https://github.com/flathub/org.learningequality.Kolibri, first. Otherwise the newer database will cause trouble with the old Kolibri in our target system.

Oh, I didn't realize it was that sensitive. Alright, let me take a look at that.

dbnicholson:
The org.learningequality.Kolibri flatpak got updated, so I believe this is good to go in now.

PR #72, opened 2022-05-19
kolibri: Use our content CDN during builds

wjt:
To offload our Learning Equality's CDN, we've setup our own CDN in front
of it at https://kolibri-content.endlessos.org. Import content using
that as the base URL. Installations are still configured to use the
Kolibri default URL. The configuration can be cleared to go back to the
default URL for builds, too.

https://phabricator.endlessm.com/T30950
https://phabricator.endlessm.com/T33499

This backports #65 to eos4.0.

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -253,6 +253,11 @@ automatic_provision_learner_can_edit_password = false
 automatic_provision_learner_can_sign_up = false
 automatic_provision_allow_guest_access = false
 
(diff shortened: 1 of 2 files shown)


dbnicholson:
I'd like to wait until #71 goes in before putting this on the stable branch. I'm worried that without it we'll build images without fully provisioned channels.

dbnicholson (approved):

PR #83, opened 2022-09-07
Temporarily install no apps or runtimes from Flathub

wjt:
Flathub removed aarch64 refs from the main summary (which hit the 10MB
limit). They are still accessible in the "sub-summary" which (lib)flatpak
knows how to use. Sadly eibflatpak.py bypasses libflatpak and attempts
to use the libostree API directly to pull Flatpak refs; libostree does
not know about sub-summaries so this fails.

### DIFF
--- a/config/arch/arm64.ini
+++ b/config/arch/arm64.ini
@@ -7,6 +7,11 @@ enable = false
 boot_zip = false
 iso = false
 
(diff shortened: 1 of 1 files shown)


wjt:
Testing this at https://ci.endlessos.org/job/image-build-arm64/4095

starnight (approved):

PR #89, opened 2022-10-20
Use Flatpak transactions

dbnicholson:
The custom Flatpak installation code had become broken for arm64. This changes it to be a bit less bespoke by relying on Flatpak transactions. That all happens in the last commit, which is massive. Apologies. While here a pile of test code was added.

### DIFF
--- a/lib/eibflatpak.py
+++ b/lib/eibflatpak.py
@@ -22,6 +22,7 @@ import base64
 import codecs
 from collections import namedtuple, OrderedDict
 from configparser import ConfigParser
+from contextlib import contextmanager
 import eib
 from eibostree import fetch_remote_collection_id
 import fnmatch
@@ -34,7 +35,7 @@ from urllib.request import urlopen
 
 require_version('Flatpak', '1.0')
 require_version('OSTree', '1.0')
-from gi.repository import Flatpak, GLib, OSTree  # noqa: E402
+from gi.repository import Flatpak, GLib, Gio, OSTree  # noqa: E402
 
 logger = logging.getLogger(__name__)
 
@@ -112,20 +113,6 @@ class FlatpakFullRef(namedtuple('FlatpakFullRef', (
     @property
     def has_extra_data(self):
         return 'Extra Data' in self.metadata
-
-
-class FlatpakInstallRef(object):
-    """Flatpak ref for installation
-
-    A wrapper around FlatpakFullRef containing a mutable subpaths
-    attribute. This is needed because the subpaths get accessed from the
-    dependent flatpak's RelatedRefs.
-    """
-    def __init__(self, full_ref):
-        self.full_ref = full_ref
-
-        # Provided later when resolving install set
-        self.subpaths = None
 
 
 class FlatpakRemote(object):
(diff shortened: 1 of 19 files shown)


dbnicholson:
I believe this is working correctly and I ran an [arm64 test build](https://ci.endlessos.org/view/Image/job/image-build-arm64/4366/) that succeeded installing flatpaks.

starnight (approved):
LGTM!

lib/eibflatpak.py
```diff
@@ -641,54 +628,73 @@ def _set_extra_languages(self):
...
+        try:
+            repo_config.remove_key('core', 'xa.masked')
+        except GLib.Error as err:
+            # Ignore errors for missing group or key
+            if err.matches(GLib.KeyFile.error_quark(),
+                           GLib.KeyFileError.GROUP_NOT_FOUND):
+                pass
+            elif err.matches(GLib.KeyFile.error_quark(),
+                             GLib.KeyFileError.KEY_NOT_FOUND):
+                pass
```
wjt, line 670: This is the WORST thing about this GLib API.

lib/eibflatpak.py
```diff
@@ -704,173 +710,195 @@ def _log_installation_free_space(self):
...
-        Add the apps and runtimes required for each remote and resolve
-        all runtime and extension dependencies.
-        """
-        # Dict of FlatpakInstallRefs needed for installation keyed by
-        # the ref string.
-        self.install_refs = {}
-        eol_refs = []
-        eol_rebase_refs = []
+    def _check_excluded_operations(self, operations):
+        """Verify none of the refs in the opertions are excluded"""
```
wjt, line 726: typo: operations

lib/eibflatpak.py
```diff
@@ -535,7 +535,7 @@ def __init__(self, installation, config=None, is_cache_repo=False):
 
         # Get locales configuration from generic flatpak section
         self.locales = self.config.get('flatpak', 'locales',
-                                       fallback=()).split()
+                                       fallback=[]).split()
```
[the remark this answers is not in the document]
starnight: oops!  I thought the list works.  After test manually, seems my understanding is wrong.
```
>>> [].split()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'list' object has no attribute 'split'
```

lib/eibflatpak.py
```diff
@@ -883,103 +911,61 @@ def _do_pull(self, repo, remote, options):
...
+    def seed(self, cache_repo_path, refs):
+        """Pull commits from a cache repo to the installation repo
 
-        Use OSTree to pull all the needed refs to a repository. If
-        commit_only is True, the commit checksums are pulled without
-        creating repository refs. If cache_repo_path points to an ostree
-        repo, it will be used as a local object cache.
+        This is used during install to get as many commit objects as
+        possible into the installation repo ahead of time. It would be
+        better if you could specify the OSTree pull localcache-repos
```
wjt, line 919: boooo

wjt (commented):
Looks great!

lib/eibflatpak.py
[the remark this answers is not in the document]
dbnicholson: Oh my! What a moron. I hit this when I was initially putting together the tests and wasn't specifying it. But I guess by the time I added the "fix" I was setting it and never actually checked that this worked. I'll follow up to fix this.

tests/util.py
```diff
@@ -26,3 +31,17 @@ def import_script(name, script):
     module = importlib.util.module_from_spec(spec)
     spec.loader.exec_module(module)
     return module
+
+
+# shlex.join added in python 3.8.
```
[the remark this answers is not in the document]
dbnicholson: It's so I could keep using ubuntu-18.04 in the GitHub tests where python is 3.6. I don't recall what we have on 4.0 but I think it's newer than that. I wanted to keep supporting that if it wasn't too much effort since we tell people they can run the image builder on basically any debian style host.

However, in a real image build this code runs in the buildroot against EOS, so it's all a bit silly. For a bit I did attempt to generate a buildroot and then run pytest inside that, but it was too complicated. If we ever make the eos containers public, I'd use those in the GitHub tests but even then I'd attempt to use the appropriate branch. So, probably we could tear this and the other compat code out at some point.

tests/eib/test_eibflatpak.py
```diff
@@ -0,0 +1,554 @@
...
+    path = tmp_path / 'local-flatpak-inst'
+    path.mkdir()
+    return path
+
+
+@pytest.fixture
+def local_flatpak_installation(local_flatpak_installation_path):
+    # import gi
+    # gi.require_version('Flatpak', '1.0')
+    # from gi.repository import Flatpak, Gio
```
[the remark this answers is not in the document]
dbnicholson: Oops, thanks. I thought I got all of those.

lib/eibflatpak.py
dbnicholson: In C it's not too bad. In GI bindings it's terrible.

lib/eibflatpak.py
dbnicholson: 👍

lib/eibflatpak.py
dbnicholson: I opened https://github.com/flatpak/flatpak/issues/1284 a while back. I think with the transaction code it would be easier to narrow that to just adding a `flatpak_transaction_set_localcache_repos` API. That feature is already used internally when pulling to a user repo and using the system repo as a cache (I think).

lib/eibflatpak.py
dbnicholson: See #90.

tests/eib/test_eibflatpak.py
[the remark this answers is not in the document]
dbnicholson: See #90.

PR #91, opened 2022-10-21
Flatpak transaction backport

dbnicholson:
I decided against doing a narrower backported and just took all the commits from #89 and #90. There were 2 issues:

* Some of the giant rewrite commit (0dae68c5f55adf120ae2c2b4ffdad742dd258b4b) didn't apply cleanly because on master I'd done some linting fixes that weren't backported. I left out the linting fixes since they're not necessary and the eos4.0 test suite isn't running `flake8`.
* The eibflatpak tests were exercising 9e0032985c69deb5f684c08fea26d9a4f73f22b1, but that wasn't backported. I just removed the parts that were trying to handle that.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,9 +10,12 @@ on:
 jobs:
   tests:
     strategy:
+      # Let other configurations continue if one fails.
+      fail-fast: false
       matrix:
         os:
           - ubuntu-20.04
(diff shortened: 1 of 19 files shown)


dbnicholson:
The test image build succeeded and there were no differences in the flatpaks installed from the previous build. I think this is ready to go.

PR #92, opened 2022-11-09
config: drop Rhythmbox from default config

starnight:
EOS is going to switch from Rhythmbox to GNOME Music as default music player. So, drop Rhythmbox from the default config.

https://phabricator.endlessm.com/T34075


wjt (changes requested):
No, I am not proposing we drop Rhythmbox from most images. We should move it from the mandatory list to the default list below. This will cause it to remain in full images and be removed from base.

wjt (approved):

wjt:
I reworded the commit message a bit.

PR #94, opened 2022-11-23
Use host ostree as localcache repo

wjt:
When running the image builder on Endless OS, there is a high chance that the host system will be running a similar, if not identical, OSTree commit to the one being built into the image. In my case, I run master and update essentially every morning, so running the image builder with its default arguments will use the same OSTree as is booted.

`ostree pull` supports a `--localcache-repo PATH` argument:

> Like git's clone --reference. Reuse the provided OSTree repo as a
> local object cache when doing HTTP fetches.

In this patch, we bind-mount the host's /ostree/repo into the buildroot, if it exists. The bind-mount is made read-only to reduce the risk that a bug in the image builder will mess up the host system.

This is obviously a hack but I thought I'd give it a try and it seems to make the build run somewhat faster locally. Of course from the second build on any given day it has no effect because the image builder's cache has the OS cached, so maybe it is not worthwhile. So this PR is more of an RFC.

### DIFF
--- a/run-build
+++ b/run-build
@@ -95,11 +95,23 @@ class ImageBuildRoot(object):
         repo_path = config['ostree']['repodir']
         remote = config['ostree']['remote']
         ref = config['ostree']['ref']
+
+        # If we are on an OSTree system, use the system repo as an additional cache.
+        # In the case of an Endless OS developer system booted into the same or similar
+        # ref as we are building an image of, this saves pulling over the network. In
+        # other cases it is harmless.
+        system_ostree_repo = '/ostree/repo'
+        if os.path.isdir(system_ostree_repo):
+            extra_pull_args = ['--localcache-repo', system_ostree_repo]
+        else:
+            extra_pull_args = []
+
         log.info(f'Pulling OSTree ref {remote}:{ref}')
         eib.retry(subprocess.check_call, [
             'ostree',
             f'--repo={repo_path}',
             'pull',
+            *extra_pull_args,
             remote,
             ref,
         ])


jprvita:
I have not looked at the code too closely, but I think this is a good idea. IIUC the image builder's cache you are referring to is per-worker, so that one only helps on the second build running on the same worker on a given day.

wjt:
I am not considering the case where this runs on Jenkins – only for local development. This will not make any difference on Jenkins because those workers are not running an OSTree version of Endless OS and so do not have `/ostree/repo`.

dbnicholson:
> @dbnicholson I remembered this patch today because I found myself waiting on the "pull the ostree" stage of an image build during development. WDYT?

I think it's a good idea, although I have a patch to construct the buildroot with an ostree (to avoid needing to do the whole thing with `mmdebstrap`). In that case the ostree is pulled right in the beginning using the host's ostree and you wouldn't need to plumb the host repo into the buildroot.

run-build
```diff
@@ -206,6 +207,8 @@ class ImageBuildRoot(object):
             with open(mount_path, 'w'):
                 pass
         subprocess.check_call(['mount', '--bind', path, mount_path])
+        if read_only:
+            subprocess.check_call(['mount', '-o', 'remount,bind,ro', path, mount_path])
```
dbnicholson, line 211: Oh, I could use this a bit.

dbnicholson:
> > @dbnicholson I remembered this patch today because I found myself waiting on the "pull the ostree" stage of an image build during development. WDYT?
> 
> I think it's a good idea, although I have a patch to construct the buildroot with an ostree (to avoid needing to do the whole thing with `mmdebstrap`). In that case the ostree is pulled right in the beginning using the host's ostree and you wouldn't need to plumb the host repo into the buildroot.

Needs a bit of polish (and a proper issue to track), but 6cd8a232fd31adf268948a5f701649e8082278f5 is what I'm talking about.

wjt:
Triggered https://ci.endlessos.org/job/image-build-amd64/28092/console to test this on a non-OSTree system; I'll kill it once it successfully pulls the OS.

wjt:
I left my system running a test build of #135 and after 35 minutes it had not yet finished pulling the master ref for its buildroot. This branch seems to work fine both on my laptop and (when I ran a test build above after rebasing on your changes to use ostree for the buildroot) on the build server. WDYT?

run-build
```diff
@@ -95,11 +95,23 @@ class ImageBuildRoot(object):
...
+            extra_pull_args = ['--localcache-repo', system_ostree_repo]
+        else:
+            extra_pull_args = []
+
         log.info(f'Pulling OSTree ref {remote}:{ref}')
         eib.retry(subprocess.check_call, [
             'ostree',
             f'--repo={repo_path}',
             'pull',
+            *extra_pull_args,
```
dbnicholson, line 114: Does the list within the list get flattened? I wouldn't expect it, but maybe subprocess does that and I've never tried.

wjt: That's what the `*` does:

```
>>> extra_pull_args = ['--localcache-repo', "/ostree/repo"]
>>> ["ostree", "pull", *extra_pull_args, "eos", "my-cool-ref"]
['ostree', 'pull', '--localcache-repo', '/ostree/repo', 'eos', 'my-cool-ref']
>>> ["ostree", "pull", extra_pull_args, "eos", "my-cool-ref"]
['ostree', 'pull', ['--localcache-repo', '/ostree/repo'], 'eos', 'my-cool-ref']
```

dbnicholson: Oh, I totally missed that. :+1:

dbnicholson: I had no idea you could compose lists that way. That's a nice feature.

dbnicholson (approved):

PR #95, opened 2022-11-29
image: Support XKB keyboard variants

GeorgesStavracas:
This will be used by some custom images.

https://phabricator.endlessm.com/T33987

### DIFF
--- a/hooks/image/50-xkb-layout.chroot
+++ b/hooks/image/50-xkb-layout.chroot
@@ -10,6 +10,15 @@ if [ -n "${EIB_IMAGE_XKB_LAYOUT}" ]; then
 		Identifier "system-keyboard"
 		MatchIsKeyboard "on"
 		Option "XkbLayout" "${EIB_IMAGE_XKB_LAYOUT}"
+	EOF
+
+	if [ -n "${EIB_IMAGE_XKB_VARIANT}" ]; then
+		cat <<- EOF >> ${TARGET_PATH}/00-keyboard.conf
+			Option "XkbVariant" "${EIB_IMAGE_XKB_VARIANT}"
+		EOF
+	fi
+
+	cat <<- EOF >> ${TARGET_PATH}/00-keyboard.conf
 	EndSection
 	EOF
 fi
--- a/hooks/image/50-xkb-layout.chroot
+++ b/hooks/image/50-xkb-layout.chroot
@@ -1,6 +1,7 @@
 #Customize the standard keyboard layout
 
 TARGET_PATH=/etc/X11/xorg.conf.d
+DEBIAN_DEFAULTS_PATH=/etc/default
 
 if [ -n "${EIB_IMAGE_XKB_LAYOUT}" ]; then
 	mkdir -p ${TARGET_PATH}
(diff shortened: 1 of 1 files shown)


hooks/image/50-xkb-layout.chroot
```diff
@@ -10,6 +10,15 @@ if [ -n "${EIB_IMAGE_XKB_LAYOUT}" ]; then
 		Identifier "system-keyboard"
 		MatchIsKeyboard "on"
 		Option "XkbLayout" "${EIB_IMAGE_XKB_LAYOUT}"
+	EOF
+
+	if [ -n "${EIB_IMAGE_XKB_VARIANT}" ]; then
+		cat <<- EOF > ${TARGET_PATH}/00-keyboard.conf
```
dbnicholson, line 16: You have to append to the file with `>>` rather than `>`.

hooks/image/50-xkb-layout.chroot
```diff
@@ -10,6 +10,15 @@ if [ -n "${EIB_IMAGE_XKB_LAYOUT}" ]; then
...
+			Option "XkbVariant" "${EIB_IMAGE_XKB_VARIANT}"
+		EOF
+	fi
+
+	cat <<- EOF > ${TARGET_PATH}/00-keyboard.conf
```
dbnicholson, line 21: Similarly here.

dbnicholson (changes requested):

GeorgesStavracas:
Okay, can confirm this works

GeorgesStavracas:
Now I can confirm this works (as in, adds the right text to the right files) AND it works (as in, it makes the OS start with the specified keyboard layout and variant)

hooks/image/50-xkb-layout.chroot
```diff
@@ -21,4 +22,17 @@ if [ -n "${EIB_IMAGE_XKB_LAYOUT}" ]; then
 	cat <<- EOF >> ${TARGET_PATH}/00-keyboard.conf
 	EndSection
 	EOF
+
+	# Write to /etc/default/keyboard too, Debian patches
+	# localed to read and write from this location
+
+	cat <<- EOF >> ${DEBIAN_DEFAULTS_PATH}/keyboard
+		XKBLAYOUT=${EIB_IMAGE_XKB_LAYOUT}
+	EOF
```
dbnicholson, line 31: Very nitpicky, but the `<<-` will strip the indent up to where the `EOF` is. That would cause `XKBLAYOUT` and `XKBVARIANT` to be indented 2 spaces, which I don't think was intended.

dbnicholson (approved):
Looks good, but I did leave one nitpick. Feel free to address that or not. Does adjusting `00-keyboard.conf` even have any effect?

hooks/image/50-xkb-layout.chroot
GeorgesStavracas: Indeed, fixed!

GeorgesStavracas:
> Does adjusting 00-keyboard.conf even have any effect?

It does, on X11 sessions. They're not default anymore, but I left it there because users might legitimately want switch back to X11, maybe even be instructed by our support, so I guess it wouldn't hurt.

dbnicholson:
> > Does adjusting 00-keyboard.conf even have any effect?
> 
> It does, on X11 sessions. They're not default anymore, but I left it there because users might legitimately want switch back to X11, maybe even be instructed by our support, so I guess it wouldn't hurt.

Interesting. I asked this in Slack, but it seems like if Debian patches localed to work with `/etc/default/keyboard`, then they must also patch X to work with it, too. Otherwise they just made localed much less useful with that change.

dbnicholson:
> > > Does adjusting 00-keyboard.conf even have any effect?
> > 
> > 
> > It does, on X11 sessions. They're not default anymore, but I left it there because users might legitimately want switch back to X11, maybe even be instructed by our support, so I guess it wouldn't hurt.
> 
> Interesting. I asked this in Slack, but it seems like if Debian patches localed to work with `/etc/default/keyboard`, then they must also patch X to work with it, too. Otherwise they just made localed much less useful with that change.

Or, to put it another way, does adjusting `00-keyboard.conf` have any effect in the presence of `/etc/default/keyboard`? I think you should just go ahead with it this way (even if we're not actually using Xorg anymore), but a question for another day.

PR #96, opened 2022-11-30
config: Bump the default stable_branch to latest2

starnight:
* Add eos4.0 config to set eos4.0 branch's stable_branch
* Bump the default stable_branch to latest2

https://phabricator.endlessm.com/T33353

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -319,7 +319,7 @@ remote = eos
 # For LTS builds this may be the built branch (e.g., eos4.0) or the
 # major series branch (e.g., eos4) to keep the system on a specific EOS
 # version.
-stable_branch = latest1
+stable_branch = latest2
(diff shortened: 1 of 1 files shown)


config/branch/eos4.0.ini
```diff
@@ -0,0 +1,8 @@
+# eos4.0 branch settings
+
+[ostree]
+# A different deploy ref is needed for eos3.4+ so that users are
+# required to get a specific eos-updater on the eos3.3 branch.
+#
+# https://phabricator.endlessm.com/T22056
+stable_branch = latest1
```
dbnicholson, line 8: Since eos4.0 builds are already running from the eos4.0 branch of this repo, I don't think we need to bother with configuration (even though it's nice documentation).

dbnicholson (approved):
Good catch. I'm going to remove the eos4.0 change as it's not relevant on this branch and then merge.

PR #98, opened 2022-12-01
config: Add xkb_variant to defaults.ini (eos5.0)

GeorgesStavracas:
Doesn't make a difference, but it's good to list it there.


wjt:
I'll just fast-forward to include #97. master and eos5.0 are level.

PR #99, opened 2022-12-01
Backport T33987 to eos4.0

GeorgesStavracas:
https://phabricator.endlessm.com/T33987

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -200,6 +200,7 @@ language =
 
 # Default Keyboard Layout
 xkb_layout =
+xkb_variant =
 
 # Default timezone
 timezone =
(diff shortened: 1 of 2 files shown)


wjt:
Don't you want 26315c2661cf0b3b860067a53a68026a0ceab030 as well?

wjt:
I have an eos4.0 VM where I configured the keyboard as `gb+dvorak` during initial setup, and `/etc/default/keyboard` there is also the only reference to `dvorak` in any config I can find, and it's used at the login screen, so I am confident that it is the file that's used on eos4.0 too.

GeorgesStavracas:
> Don't you want https://github.com/endlessm/eos-image-builder/commit/26315c2661cf0b3b860067a53a68026a0ceab030 as well?

Of course. Not sure why I didn't pick that commit

PR #105, opened 2023-02-15
buildscript: Declare repo as a git safe directory

dbnicholson:
Newer git prevents operations when the calling user is not the same as the user owning the repo. Currently the only git operation run is `git rev-parse HEAD` to record the builder commit in the manifest, but even that is enough to make git fail. Add the repo as a safe directory inside the build root since the builder is running as root and the repo may not be owned by root.

https://community.endlessos.com/t/custom-build-problem-with-5-0/19900

### DIFF
--- a/buildscript
+++ b/buildscript
@@ -21,6 +21,10 @@ done
 # Change back to the source directory since we've chrooted.
 cd "${EIB_SRCDIR}"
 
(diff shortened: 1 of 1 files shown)


jprvita (approved):
This looks like it will do the right thing (although I haven't tested it). Feel free to merge.

PR #108, opened 2023-02-21
config: Add Epiphany to all images

dbnicholson:
Although Chromium remains the default browser, Epiphany is required for Personal Web Application (PWA) support.

https://phabricator.endlessm.com/T34479


wjt (approved):
Same nitpick: P is for Progressive.

dbnicholson:
Updated the commit message and added a commit to drop the ubuntu-18.04 runner since it's deprecated and currently in one of it's forced fail times.

PR #111, opened 2023-03-03
Kolibri remote import

dbnicholson:
If a custom Kolibri content server is in use and credentials are
available, use the API to import the desired channels remotely. When the
time comes to populate the channels on the image, all the content will
be available on the server.

https://phabricator.endlessm.com/T33705

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,6 +14,6 @@ indent_style = space
 indent_size = 2
 
 # Python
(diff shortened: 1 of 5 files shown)


dbnicholson:
> Very neat.
> 
> I'm not super familiar with the Kolibri API (to say the least). Still I'll go ahead and merge it because an end to end test seems like the best way to see how it works in practice. @dylanmccall maybe you'll have some comments for a follow-up

Indeed, I'm not even sure they really intend the API to be used remotely like this. It's primarily used between the frontend JS and the backend Django app. However, since we contol the server this is talking to, we shouldn't be surprised if the API changes. Hmm, that makes me think I should be restricting the `kolibri` version in the server build.

Anyways, I did test this locally by putting the credentials in `/etc/eos-image-builder/netrc` and configuring a build to import my favorite tiny channel - Passionfruit Island.

wjt:
It's not listed on my favourite Wikipedia page, List of Fictional Islands.

https://en.wikipedia.org/wiki/List_of_fictional_islands#P

Alas

PR #112, opened 2023-03-20
ostree: Don't verify summary on HTTPS

dbnicholson:
Don't bother GPG verifying the summary if it's being fetched over HTTPS. This often fails when the summary is being updated concurrently since the received summary and summary signature can be mismatched. The built image is already configured to not verify the summary.

https://phabricator.endlessm.com/T32658

### DIFF
--- a/stages/eib_ostree
+++ b/stages/eib_ostree
@@ -21,11 +21,19 @@ else
 fi
 
 # Pull in the latest commit. Recreate the remote setup to ensure there
-# aren't any stale settings.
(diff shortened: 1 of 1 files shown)


wjt (approved):
I had missed that we do this in the OS. Has it always been that way and I have just not noticed??

dbnicholson:
> I had missed that we do this in the OS. Has it always been that way and I have just not noticed??

Yes, I think. Ostree by default doesn't verify summaries. If you add a remote from flatpak it does, though.

PR #113, opened 2023-03-21
gnome-software-cache: Update eos-extra cached appstream path

dbnicholson:
2 changes are needed:

* gnome-software now uses the non-legacy `/var/cache/swcatalog/xml` AppStream directory.
* gnome-software now includes a SHA1 hash of URL in the filename in case there are multiple external AppStream URLs with the same basename.

https://phabricator.endlessm.com/T34337


dbnicholson:
I changed my mind on this. Since `eos-extra.xml` is now part of a translation pipeline, it would be better if it was not checked into version control. However, that means the file that's being copied here won't exist. Instead, it should be downloaded. Which is also what would happen if we can get `gnome-software-cmd refresh` working.

PR #114, opened 2023-03-21
gnome-software-cache: Update eos-extra cached appstream path

dbnicholson:
2 changes are needed for the path:

* gnome-software now uses the non-legacy `/var/cache/swcatalog/xml` AppStream directory.
* gnome-software now includes a SHA1 hash of URL in the filename in case there are multiple external AppStream URLs with the same basename.

This also switches the XML to being downloaded rather than copied from the git repo. This validates the URL and ensures that the exact file is installed. This is what would happen if we later switch to using `gnome-software-cmd refresh`, anyways.

https://phabricator.endlessm.com/T34337


pwithnall (approved):
Looks reasonable to me

PR #119, opened 2023-09-06
Create separate KOLIBRI_HOME for Endless Key

jprvita:
We want to be able to ship different content and configuration for
Kolibri and the Endless Key app. This commit adds support for installing
the content and configuration for each app to separate locations, so
they can be used side-by-side.

https://github.com/endlessm/eos-image-builder/issues/117

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -307,10 +307,6 @@ regular_users_can_manage_content = false
 # listed in this setting.
 install_channels =
 
-# By default we provision the Endless Key facility with the same settings from
-# the [kolibri] section.
-automatic_provision = true
-
 [usb]
 size = 16000000000
 free_space = 1000
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -301,6 +301,16 @@ regular_users_can_manage_content = false
 #   # Kolibri 0.12.2 User Guide for Admins [document]
 #   5bb37c1832c8489ab2940f31588305f6
 
+[endlesskey]
+# Preloaded content or the Endless Key app
+# The Endless Key app can be pre-loaded with content from the Kolibri channels
+# listed in this setting.
+install_channels =
+
+# By default we provision the Endless Key facility with the same settings from
+# the [kolibri] section.
+automatic_provision = true
+
 [usb]
 size = 16000000000
 free_space = 1000
--- a/hooks/image/60-kolibri-content
+++ b/hooks/image/60-kolibri-content
@@ -1,6 +1,7 @@
 # Populate the Kolibri home directory
 
-if [ -z "${EIB_KOLIBRI_INSTALL_CHANNELS}" ]; then
+if [ -z "${EIB_KOLIBRI_INSTALL_CHANNELS}" ] &&
+   [ -z "${EIB_ENDLESSKEY_INSTALL_CHANNELS}" ]; then
   exit 0
 fi
 
@@ -15,9 +16,19 @@ export KOLIBRI_STATIC_USE_SYMLINKS=0
 
 import_kolibri_channel()
 {
-  local channel_id=$1
-  local channel_include_node_ids_var="EIB_KOLIBRI_${channel_id^^}_INCLUDE_NODE_IDS"
-  local channel_exclude_node_ids_var="EIB_KOLIBRI_${channel_id^^}_EXCLUDE_NODE_IDS"
+  local app_name=$1
+  local channel_id=$2
+
+  if [ ${app_name} == "kolibri" ]; then
+    local channel_include_node_ids_var="EIB_KOLIBRI_${channel_id^^}_INCLUDE_NODE_IDS"
+    local channel_exclude_node_ids_var="EIB_KOLIBRI_${channel_id^^}_EXCLUDE_NODE_IDS"
+  elif [ ${app_name} == "endless-key" ]; then
+    local channel_include_node_ids_var="EIB_ENDLESSKEY_${channel_id^^}_INCLUDE_NODE_IDS"
+    local channel_exclude_node_ids_var="EIB_ENDLESSKEY_${channel_id^^}_EXCLUDE_NODE_IDS"
+  else
+    exit 1
+  fi
+
   local importcontent_opts=(
     # Normally importcontent ignores download errors. Make it fail so we
     # can be sure we've fully provisioned channels.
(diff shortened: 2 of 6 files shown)


hooks/image/60-kolibri-content
```diff
@@ -45,42 +66,122 @@ import_kolibri_channel()
...
+    PLUGINS_DISABLE=(
+      "kolibri.plugins.default_theme"
+      "kolibri.plugins.learn"
+    )
+    PLUGINS_ENABLE=(
+      "kolibri.plugins.app"
+      "kolibri_endless_key_theme"
+      # FIXME: enabling kolibri-explore-plugin fails because we are using a
+      #        kolibri version that is too old.
+      #"kolibri_explore_plugin"
```
dylanmccall, line 109: I think this is fine. It means Kolibri will need to execute some of extra database migrations on first run, but shouldn't be a problem.

One issue we are going to crash into here is that the Explore plugin provides its own [`applyexternaltags`](https://github.com/endlessm/kolibri-explore-plugin/blob/master/kolibri_explore_plugin/management/commands/applyexternaltags.py) management command. We use that when importing content from our content collection files, and it's necessary so the Explore plugin knows how to populate the different sections in the Discovery page. Without those tags, the Discovery page is pretty bare.

I think that is going to be a bit of a puzzle to sort out, so I don't know if we want to do it _here_, but we'll want to do this all a bit differently in the near future. The good news is, I don't think we will need to put a Kolibri database in the image at all any more, so this issue about the Explore plugin and the old Kolibri version will probably be moot. All we need is to fetch the content files, put them somewhere (as in `kolibri manage exportcontent`), add a JSON file describing what that content is (the `exportcontent` command generates a content manifest file), and get the app to populate its database (with the correct content in the database and marked available) based on that file.

For Kolibri, that JSON file would be exactly what the `exportcontent` command generates. For Endless Key, we'd copy our own JSON files from [endless-key-collections](https://github.com/endlessm/endless-key-collections). Those ones include the information we need about external content tags.

config/defaults.ini
```diff
@@ -301,6 +301,16 @@ regular_users_can_manage_content = false
 #   # Kolibri 0.12.2 User Guide for Admins [document]
 #   5bb37c1832c8489ab2940f31588305f6
 
+[endlesskey]
+# Preloaded content or the Endless Key app
+# The Endless Key app can be pre-loaded with content from the Kolibri channels
+# listed in this setting.
+install_channels =
```
dylanmccall, line 308: Is it intentional for the moment that we're going to do the old content nodes exhaustively listed in the image configuration? Another option here is we could add an `install_content_collections` option which lists names of collection files from [endless-key-collections](https://github.com/endlessm/endless-key-collections). The good news about those is Kolibri's `importcontent` management command already knows how to read them. The bad news is it would involve a whole new hook (and probably a Python script because that management command is a smidge limited at the moment).

dylanmccall (changes requested):
I have some thoughts here, but I'm not sure how urgently we want to address them.

The one thing I definitely think we should change is we don't need these changes for the `image/62-kolibri-automatic-provision` hook to support Endless Key. That stuff just happens on its own with the current version of the Endless Key app, and I don't think we'll have images needing it to behave differently.

hooks/image/60-kolibri-content
dbnicholson: It's certainly a lot simpler to put the content directly in the Kolibri homedir rather than putting it somewhere else on disk and then coming up with a procedure to import it and then maybe delete the massive duplicated data.

When I was previously thinking about this, my understanding was that the content pack would still be picked at runtime. Then the explore plugin would very quickly run through the content downloader since all the content is already available. The last step would to apply the tags per the chosen content pack would still run.

hooks/image/60-kolibri-content
```diff
@@ -143,4 +153,35 @@ fi
...
+    jq -r '.assets[] | select(.name == "apps-bundle.zip") | .browser_download_url'
+  )
+  collections_url=$(
+    curl -sSL https://api.github.com/repos/endlessm/endless-key-collections/releases/latest |
+    jq -r '.tarball_url'
+  )
+
+  PLUGIN_LOCATION=`pip3 show kolibri-explore-plugin | grep Location  | awk '{ print $2 }'`
+  APPS_LOCATION="${PLUGIN_LOCATION}/kolibri_explore_plugin"
+  COLLECTIONS_LOCATION="${PLUGIN_LOCATION}/kolibri_explore_plugin/static/collections/"
```
dbnicholson, line 167: I don't think we need to do this. These bundles are only needed at runtime and aren't required if you're just running `importchannel`/`importcontent`.

dbnicholson (changes requested):
Looks like this is getting closer, but I think it's going to need a bit more work.

config/defaults.ini
```diff
@@ -301,6 +301,16 @@ regular_users_can_manage_content = false
...
+
+# By default we provision the Endless Key facility with the same settings from
+# the [kolibri] section.
+automatic_provision = true
+
```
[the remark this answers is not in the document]
jprvita: I had already dropped the code that uses it, but left the option behind. Thanks!

hooks/image/62-kolibri-automatic-provision
```diff
@@ -27,3 +26,13 @@ cat <<EOF > "${OSTREE_VAR}"/lib/kolibri/data/automatic_provision.json
...
+
+if [ "${EIB_KOLIBRI_AUTOMATIC_PROVISION}" == "true" ]; then
+  mkdir -p "${OSTREE_VAR}"/lib/kolibri/data
+  cp "${EIB_TMPDIR}"/automatic_provision.json "${OSTREE_VAR}"/lib/kolibri/data
+fi
+
+if [ "${EIB_ENDLESSKEY_AUTOMATIC_PROVISION}" == "true" ]; then
+  mkdir -p "${OSTREE_VAR}"/lib/endless-key/data
+  cp "${EIB_TMPDIR}"/automatic_provision.json "${OSTREE_VAR}"/lib/endless-key/data
+fi
```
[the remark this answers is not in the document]
jprvita: Already addressed.

hooks/image/60-kolibri-content
jprvita: But don't we need to put them in place, so they can be found at runtime? Or do they already come with the Flatpak?

hooks/image/60-kolibri-content
```diff
@@ -45,42 +56,62 @@ import_kolibri_channel()
...
+  local app_name=$1
+  # Needs to be kept in sync with hooks/image/61-kolibri-content-install
+  KOLIBRI_CONTENT_DIR="${EIB_CONTENTDIR}/${app_name}-content"
+  # FIXME: For now, we need to remove old content that may exist from previous
+  # runs to prevent images accidentally getting extra channels that are not in
+  # their configuration. However, this defeats the purpose of the image builder
+  # saving the contents of EIB_CONTENTDIR between runs to improve its performance.
+  # At some point we should revisit this to try to make the caching work properly.
+  rm -rf "${KOLIBRI_CONTENT_DIR}"
+  mkdir -p "${KOLIBRI_CONTENT_DIR}"
```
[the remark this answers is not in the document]
jprvita: Agreed.

hooks/image/60-kolibri-content
dbnicholson: The flatpak has them. Otherwise, how would it work if you just install the flatpak?

https://github.com/flathub/org.endlessos.Key/blob/master/modules/endless-key-collections.json
https://github.com/flathub/org.endlessos.Key/blob/master/modules/python3-kolibri-explore-plugin.json

jprvita: That makes sense.. I was confused [by your earlier comment](https://github.com/endlessm/eos-image-builder/issues/117#issuecomment-1688784247), but I guess things were not as well defined at that point.

PR #120, opened 2023-09-27
Preload Endless Key content

jprvita:
We want to preload the whole Endless Key library in Endless OS images.
We also need to ensure the content is available in our Kolibri cache
server, if we are pulling the content from it.

https://github.com/endlessm/eos-image-builder/issues/117

### DIFF
--- a/hooks/image/51-ek-content-list
+++ b/hooks/image/51-ek-content-list
@@ -1,0 +1,15 @@
+# Populate the Endless Key home directory
+
+if [[ ! "${EIB_FLATPAK_REMOTE_FLATHUB_APPS}" =~ .*"org.endlessos.Key".* ]]; then
+  exit 0
+fi
+
+# This file is used by hooks/image/52-ek-content-cache and
+# hooks/image/53-ek-content-preload
+EK_CHANNELS_FILE="${EIB_TMPDIR}"/ek-channels
+
+rm -f "${EK_CHANNELS_FILE}"
+collections="${OSTREE_VAR}"/lib/flatpak/app/org.endlessos.Key/current/active/files/share/endless-key/collections/*.json
+for collection in ${collections}; do
+  jq -r '.channels[].id' "${collection}" >> "${EK_CHANNELS_FILE}"
+done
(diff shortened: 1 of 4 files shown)


jprvita:
When reviewing, please keep in mind that there are some unrelated failures happening -- see https://github.com/endlessm/eos-image-builder/issues/117#issuecomment-1736659174 and newer comments. That said, this code will only run for builds that include the Endless Key app and explicitly say which collections should be preloaded in the image configuration, so it should not break other image builds.

hooks/image/51-ek-content-list
```diff
@@ -0,0 +1,24 @@
...
+rm -f "${EK_CHANNELS_FILE}"
+collections="${OSTREE_VAR}"/lib/flatpak/app/org.endlessos.Key/current/active/files/share/endless-key/collections/*.json
+for collection in ${collections}; do
+  # Check if the file basename stripped off of -0001.json is part of the list
+  # of collections to be installed.
+  bn=$(basename ${collection})
+  if [[ "${EIB_ENDLESSKEY_COLLECTIONS}" =~ .*"${bn%-????.json}".* ]] ; then
+    jq -r '.channels[].id' "${collection}" >> "${EK_CHANNELS_FILE}"
+  fi
+done
```
dbnicholson, line 24: It would be nice to do the sorting once here. Like:

```
rm -f "${EK_CHANNELS_FILE}" "${EK_CHANNELS_FILE}.tmp"
for collection in ${collections}; do
  some channel filtering >> "${EK_CHANNELS_FILE}.tmp"
done
if [ -f "${EK_CHANNELS_FILE}.tmp" ]; then
  sort -u "${EK_CHANNELS_FILE}.tmp" > "${EK_CHANNELS_FILE}"
fi
```

dbnicholson (approved):
I think this works. I added 2 more commits to update the Kolibri version in use and it ran successfully locally. A nice followup would be to split the content server seeding out into a separate helper that takes a list of desired channels, but that's not needed here.

hooks/image/51-ek-content-list
jprvita: Yes, good point! So we don't try to cache / import a channel more than once. I'll change that.

jprvita:
> I added 2 more commits to update the Kolibri version in use and it ran successfully locally.

Thanks, those LGTM!

PR #122, opened 2023-10-17
config: Install Endless Key by default

dbnicholson:
I think we want this installed by default for basically all images, although images that want preloaded content should also explicitly add it. For now I left org.learningequality.Kolibri preinstalled.

### DIFF
--- a/config/branch-personality/master-base.ini
+++ b/config/branch-personality/master-base.ini
@@ -1,9 +1,0 @@
-# Settings for base personality with master branch
-
-[flatpak-remote-flathub]
-apps_del =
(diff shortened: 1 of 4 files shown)


dbnicholson:
> How much hassle would it be to remove Kolibri from the base config, but readd it in all the configs where Kolibri channels are installed? (In general we are going to shift away from plain Kolibri to Endless Key as our content shipping solution)

Not too much as most of them already do. I started looking at that but punted. It will need a companion change in endless-image-config.

dbnicholson:
I added a commit to remove the default Kolibri flatpak install.

PR #123, opened 2023-11-09
hooks: Initialize Kolibri homedir before setting variables

dbnicholson:
Kolibri seems to have an unwelcome habit of persisting the values of environment variables in the options file when the homedir is being initialized. Among other things, that would cause our Kolibri content server to be used at runtime. As a bonus, all the subsequent management commands can be run with `--skip-update` since the database migration has already happened.

Helps: https://github.com/endlessm/endless-key-content-private/issues/103

### DIFF
--- a/hooks/image/53-ek-content-preload
+++ b/hooks/image/53-ek-content-preload
@@ -5,6 +5,17 @@ if [ ! -f "${EIB_TMPDIR}"/ek-channels ]; then
   exit 0
 fi
 
+venv_dir="${EIB_TMPDIR}/kolibri-content-venv"
+python3 -m venv ${venv_dir}
+source ${venv_dir}/bin/activate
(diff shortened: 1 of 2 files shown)


wjt (approved):
Looks great.

Stupid question: how do these preloaded files end up being owned by the correct Unix user, `kolibri` (uid 132)?

dbnicholson:
> Looks great.
> 
> Stupid question: how do these preloaded files end up being owned by the correct Unix user, `kolibri` (uid 132)?

It's a little sketchy, but eos-kolibri ships a [tmpfiles config](https://github.com/endlessm/eos-kolibri/blob/master/data/tmpfiles.d/eos-kolibri.conf.in) that changes ownership using the `Z` type.

wjt:
I agree, that is a little sketchy!

PR #124, opened 2023-11-09
Fetch all Kolibri content nodes

dbnicholson:
Kolibri defaults to only downloading content nodes that it believes the current instance can render. Both the image builder and the content server have Kolibri instances that don't necessarily have all available content renderers. To be safe they should download content nodes without regard for whether Kolibri thinks they're renderable. For the `importcontent` CLI, that's the `--include-unrenderable-content` option. For the task API, that's the `renderable_only` option.

Helps: https://github.com/endlessm/endless-key-content-private/issues/103

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,6 +14,6 @@ indent_style = space
 indent_size = 2
 
 # Python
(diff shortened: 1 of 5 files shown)


dylanmccall:
Yep, this makes sense.

PR #126, opened 2023-11-10
Export appstream catalogue for preinstalled Flatpak apps

wjt:
Although we have added the name and summary for each app to the JSON manifest, there is no substitute for having all the data. XML compresses very well!

It is better to capture this at build time rather than fetching it on the fly later because apps can change or disappear, but we would like to be able to describe exactly what is in any given image.

https://phabricator.endlessm.com/T35013

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -145,6 +145,7 @@ hooks_add =
   62-kolibri-automatic-provision
   62-kolibri-options
   63-icon-grid
+  70-flatpak-appstream-catalog
   70-flatpak-manifest
   70-ostree-manifest
   80-ldconfig-aux-cache.chroot
--- a/hooks/image/70-flatpak-appstream-catalog
+++ b/hooks/image/70-flatpak-appstream-catalog
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+
+# Construct AppStream XML for preinstalled apps
+
+import logging
+import os
+import subprocess
+
+from xml.etree import ElementTree
+
+import gi
+
+gi.require_version("Flatpak", "1.0")
+from gi.repository import GLib, Gio, Flatpak  # noqa: E402
+
+logging.basicConfig(
+    level=logging.DEBUG,
+    format="+ %(asctime)s %(levelname)s %(name)s: %(message)s",
+    datefmt="%H:%M:%S",
+)
+logger = logging.getLogger(os.path.basename(__file__))
+
+APPSTREAM_VERSION = "0.8"
+
+
+def main():
+    GLib.set_prgname(os.path.basename(__file__))
+
+    image_version = os.environ["EIB_OUTVERSION"]
+    appstream_filename = image_version + ".appstream.xml"
+    appstream_path = os.path.join(os.environ["EIB_OUTDIR"], appstream_filename)
+
+    catalog = ElementTree.Element(
+        "components", version=APPSTREAM_VERSION, origin=image_version
+    )
+
+    system_path = os.path.join(os.environ["OSTREE_VAR"], "lib/flatpak")
+    logger.info("Opening Flatpak installation in %s", system_path)
+    system_file = Gio.File.new_for_path(system_path)
+    system = Flatpak.Installation.new_for_path(system_file, user=False)
+
+    installed_refs_by_origin_arch = {}
+    for ref in system.list_installed_refs():
+        key = (ref.get_origin(), ref.get_arch())
+        installed_refs_by_origin_arch.setdefault(key, set()).add(ref.format_ref())
+
+    for (origin, arch), installed_refs in installed_refs_by_origin_arch.items():
+        path = os.path.join(
+            system_path, "appstream", origin, arch, "active", "appstream.xml"
+        )
+
+        try:
+            remote_catalog = ElementTree.parse(path).getroot()
+        except ElementTree.ParseError as e:
+            # TODO: Should this be fatal?
+            logger.warning("Failed to parse %s: %s", path, e)
+            continue
+
+        metainfo_version = remote_catalog.attrib["version"]
+        if metainfo_version != APPSTREAM_VERSION:
+            logger.warning(
+                "Remote %s %s has AppStream version %s, not %s",
+                origin,
+                arch,
+                metainfo_version,
+                APPSTREAM_VERSION,
+            )
+
+        logger.info("Adding components from %s", path)
+        for component in remote_catalog:
+            bundle = component.find("bundle[@type='flatpak']")
+            try:
+                installed_refs.remove(bundle.text)
+            except KeyError:
+                pass  # Not installed
+            else:
+                catalog.append(component)
+
+        # Anything left in the set is unexpectedly not present in the AppStream for the
+        # remote it came from. It is normal for runtimes to not have metainfo; for
+        # example, .Locale extensions never do.
+        for ref in installed_refs:
+            if ref.startswith("app/"):
+                # TODO: Should this be fatal?
+                logger.warning("No component found for %s from %s", ref, origin)
+
+    tree = ElementTree.ElementTree(catalog)
+    tree.write(appstream_path, encoding="unicode", xml_declaration=True)
+
+    logger.info("Compressing %s to %s.gz", appstream_path, appstream_path)
+    subprocess.run(["pigz", "-9", "-f", appstream_path], check=True)
+
+
+if __name__ == "__main__":
+    main()
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,6 +14,6 @@ indent_style = space
 indent_size = 2
 
 # Python
-[{*.py,eos-image-builder,run-build,helpers/{assemble-manifest,fetch-remote-collection-id,generate-ovf-files,kill-chroot-procs,kolibri-pick-content-from-channel,mutable-path,packages-manifest},hooks/{content/{50-flatpak,50-kolibri-content},image/{50-flatpak.chroot,52-ek-content-cache,62-kolibri-options,70-flatpak-manifest,70-ostree-manifest}}}]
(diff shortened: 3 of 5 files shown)


config/defaults.ini
```diff
@@ -88,6 +89,7 @@ packages_add =
   json-glib-tools
   jq
   libfsapfs-utils
+  librsvg2-bin
```
wjt, line 92: I don't know that the explanation in the commit message is true, but I do know that without it my test build failed, and I am running a new build before leaving for the day to test this solution…

wjt: Sadly, that did not address the issue.

```
+ 16:30:07 INFO 70-flatpak-appstream-catalog: Composing AppStream XML: appstream-compose --origin=eoscustom-master-amd64-amd64.231110-162702.base --prefix=/var/cache/eos-image-builder/tmp/ostree-co/ostree/deploy/eos/var/lib/flatpak/exports --output-dir=/var/cache/eos-image-builder/tmp/out --icons-dir=/var/cache/eos-image-builder/tmp/appstream-compose-icons org.freedesktop.Bustle
Processing application org.freedesktop.Bustle
Error loading desktop file: Failed to load icon: image format was not recognized
```

I'm pretty sure it is an SVG thing

wjt: It helps to install the correct package, which I could have cribbed from the eos-ostree-builder commit I am taking inspiration from.

hooks/image/70-flatpak-appstream-catalog
```diff
@@ -0,0 +1,88 @@
...
+    # Exit if there aren't any desktop metainfo files.
+    if len(ids) == 0:
+        logger.info('No metainfo files')
+        return
+
+    # Compose the AppStream XML.
+    # TODO: Switch to appstreamcli compose once we are on Bookworm
+    cmd = [
+        'appstream-compose',
+        f'--origin={os.environ["EIB_OUTVERSION"]}',
```
wjt, line 76: This causes the filename to be something like `eoscustom-master-amd64-amd64.231110-173131.base.xml.gz`. [According to the spec](https://www.freedesktop.org/software/appstream/docs/chap-CatalogData.html#sect-AppStream-XML) this is an appropriate filename:

> The XML files must have a unique name, which is usually the distribution's name and version, combined with the name of the repository/origin. For example in Debian 8 (Jessie), the filename for the main repository component would be `debian-jessie-main.xml.gz`. For Fedora 20 (Heisenbug) updates it would be `fedora-20-updates.xml.gz`. 3rd-party repositories use a vendor name and repository-name combination, for example Ubuntu PPAs might get `ppa-ubuntu12.04-username-foobar.xml`.

We might like to put `.catalog` or `.appstream-catalog` in the filename I guess?

wjt:
Triggered https://ci.endlessos.org/job/image-build-amd64/28044 to test this

hooks/image/70-flatpak-appstream-catalog
dbnicholson: I'd like that. Like `.appstream.xml.gz` or something. To me that's clear enough that it's a catalog since you'd use `.metainfo.xml` if it was a single component. But I don't feel super strongly either way. Only that a bare `.xml` file will seem ambiguous in the context of image build assets and not "some place where you put appstream catalogs".

hooks/image/70-flatpak-appstream-catalog
```diff
@@ -0,0 +1,88 @@
...
+        #   preinstalled app, but still)
+        #
+        # - <release> elements are not in order (FileRoller)
+        #
+        # None of these are reasons to exclude the app from the catalogue but there is
+        # no way to tell the tool to ignore them: if we pass these files to
+        # appstream-compose it will fail.
+        #
+        # TODO: Switch to appstreamcli validate-relax once we are on Bookworm and can
+        # use appstreamcli compose.
```
dbnicholson, line 47: The other way to do this is not to use compose at all but to just build an XML catalog manually like [here](https://github.com/endlessm/gnome-software-data/blob/master/app-info/generate-eos-extra-appstream). I don't know if you'll run into the same post-catalog validation problems, though. Although, in that case you can just use `appstreamcli validate-relax`. I'm pretty sure flatpak [builds the appstream catalog manually](https://github.com/flatpak/flatpak/blob/14d735bfdaf211d13b7f0b4b9f203b847a581f39/common/flatpak-utils.c#L5354) internally.

dbnicholson (commented):
I think not using compose to build the catalog is probably a good idea as we know the components are legal enough.

hooks/image/70-flatpak-appstream-catalog
wjt: I think you're right. This also avoids stuff like dealing with icons.

wjt: It turns out not to be an option to build the XML catalog from files in `/var/lib/flatpak/exports/share/metainfo` because, despite using the newer directory name `metainfo`, files in this directory (I'm looking at you, Firefox) may be in the old pre-AppStream 0.6 format where the top-level element is not a `<component>`. appstream-glib knows how to read these and migrate them to the new format.

The next stop was to try iterating through `/var/lib/flatpak/app/...../files/share/app-info/xmls` which are the individual `<component>`s that get concatenated by `flatpak build-update-repo` to make the repo appstream. But apart from having to duplicate Flatpak's logic to only export the one(s? not sure) matching the app ID, this also means you don't get the `<bundle type="flatpak">app/ID/ARCH/BRANCH</bundle>` tag, which is actually the only reliable way to map from a Flatpak ID to an entry in an AppStream catalog. These `<bundle>` tags are only added during `flatpak build-update-repo`.

And then inspiration struck: I have been approaching this from the wrong direction. `/var/lib/flatpak/appstream/$REMOTE/$ARCH/active/appstream.xml` contains a superset of the information we want; so we just have to filter those by the `<bundle>` tag.

wjt:
Triggered https://ci.endlessos.org/job/image-build-amd64/28072 to test this again

hooks/image/70-flatpak-appstream-catalog
dbnicholson: Right. I'd been hoping to avoid digging through all the remote appstream catalogs, but I guess if that's the only way...

wjt: Actually I think this is optimal because at each stage in the build process more information is added to the data for each app; so by going to the remote catalogues we are getting the most-refined form of the data!

hooks/image/70-flatpak-appstream-catalog
```diff
@@ -0,0 +1,95 @@
...
+
+        try:
+            metainfo = ElementTree.parse(path).getroot()
+        except ElementTree.ParseError as e:
+            # TODO: Should this be fatal?
+            logger.warning("Failed to parse %s: %s", path, e)
+            continue
+
+        metainfo_version = metainfo.attrib["version"]
+        if metainfo_version != APPSTREAM_VERSION:
```
dbnicholson, line 60: Unfortunately, this seems likely to fail sooner than later - https://blog.tenstral.net/2023/11/appstream-1-0-released.html. I don't think you should try harder than this right now, but eventually it will rear it's head where some remotes are publishing 0.8 and others are publishing 1.0. Although, maybe `flatpak build-update-repo` will always enforce 0.8 when generating the catalog or something like that.

hooks/image/70-flatpak-appstream-catalog
```diff
@@ -0,0 +1,95 @@
...
+        key = (ref.get_origin(), ref.get_arch())
+        installed_refs_by_origin_arch.setdefault(key, set()).add(ref.format_ref())
+
+    for (origin, arch), installed_refs in installed_refs_by_origin_arch.items():
+        path = os.path.join(
+            system_path, "appstream", origin, arch, "active", "appstream.xml"
+        )
+
+        try:
+            metainfo = ElementTree.parse(path).getroot()
```
dbnicholson, line 53: Nitpicky, but you're parsing a catalog rather than an individual component's metainfo.

hooks/image/70-flatpak-appstream-catalog
```diff
@@ -0,0 +1,95 @@
...
+        for ref in installed_refs:
+            if ref.startswith("app/"):
+                # TODO: Should this be fatal?
+                logger.warning("No component found for %s from %s", ref, origin)
+
+    tree = ElementTree.ElementTree(catalog)
+    tree.write(appstream_path, encoding="unicode", xml_declaration=True)
+
+    logger.info("Compressing %s", appstream_path)
+    subprocess.run(["pigz", "-9", "-f", appstream_path], check=True)
```
dbnicholson, line 91: If you're compressing it, shouldn't the filename be `*.appstream.xml.gz`? Oh, I see that `pigz` (like `gzip`) will delete the original uncompressed version.

dbnicholson (approved):
Seems nice. I left some minor comments and questions, but this seems good enough to go if the test build succeeds.

hooks/image/70-flatpak-appstream-catalog
wjt: I agree it's likely to fail and I don't really know what to do about it. Perhaps just delete this whole block and assume it's good enough for the time being; and revisit once Flathub makes the noisy transition to appstream 1.0?

hooks/image/70-flatpak-appstream-catalog
wjt: I guess this variable name survived from the first 3 versions of this patch :) I'll fix it tomorrow

hooks/image/70-flatpak-appstream-catalog
wjt: That's it - can I make this clearer? I could add the target filename to the log message?

wjt: I improved the log message

wjt:
I pushed one more commit to make S3 speak the truth about the file's content-type.

dbnicholson (approved):
Looks good. I agree on the `Content-Type` handling. We/I played around with setting `Content-Encoding: gzip` in the past. It think it would work correctly if you had `Content-Type: application/xml` and the filename didn't have the `.gz` suffix. The purpose of `Content-Encoding` is that the client is in charge of decoding on transfer, which isn't what you wanted here. The nice thing when you do it in S3 is that clients can request with `Accept-Encoding: gzip` and S3 doesn't have to spend time compressing it on the fly. But, again, the intention here is that the user receives a gzipped XML file.

PR #127, opened 2023-11-13
Use ostree checkout for buildroot

dbnicholson:
The buildroot construction with `mmdebstrap` is fragile and depends on knowing the apt sources for the OS branch. Instead, use the ostree that's going to be used for the image for the buildroot since that already has the proper apt configuration. That means that the host requires `ostree` rather than `mmdebstrap` and `rsync`. In 2023 I think that's reasonable.

https://phabricator.endlessm.com/T35019


dbnicholson:
I got a little worried about the buildroot checkout corrupting the ostree objects, so I added a fixup to copy the objects. It's definitely much slower, but it should be safer. It would be nice use an overlayfs or something like that.

run-build
```diff
@@ -101,49 +94,53 @@ class ImageBuildRoot(object):
...
+
+        # Checkout the commit. The checkout is going to be mutated, so copy the
+        # objects instead of hardlinking to ensure they aren't corrupted.
+        log.info(f'Checking out OSTree ref {remote}:{ref} to {self.builddir}')
+        shutil.rmtree(self.builddir)
+        subprocess.check_call([
+            'ostree',
+            f'--repo={repo_path}',
+            'checkout',
+            '--force-copy',
```
wjt, line 120: I checked the documentation for this argument and it says:

> Never hardlink (but may reflink if available)

Oh! for btrfs or xfs or some other COW-capable filesystem.

lib/eib.py
```diff
@@ -402,49 +401,27 @@ def setup_logging():
...
+    if 'localdatadir' in config['build']:
+        keyspaths.append(os.path.join(config['build']['localdatadir'], 'keys'))
+
+    keysdirs = list(filter(os.path.isdir, keyspaths))
+    if len(keysdirs) == 0:
+        raise ImageBuildError('No gpg keys directories in', ' '.join(keyspaths))
+
+    keys = sorted(itertools.chain.from_iterable(
+        glob.iglob(os.path.join(d, '*.asc')) for d in keysdirs
+    ))
```
wjt, line 420: Not a criticism of your code, but I really hate how the only two options for flattening [[T]] to [T] in a list/generator comprehension are this or 

```python
    keys = sorted(
        x
        for d in keysdirs
        for x in glob.iglob(os.path.join(d, '*.asc'))
    )
```

neither of which is particularly pleasing on the eye.

run-build
```diff
@@ -101,49 +94,53 @@ class ImageBuildRoot(object):
...
-            '--variant=minbase',
-            '--include=' + ','.join(packages),
-            '--setup-hook=' + setup_script,
-            '--customize-hook=' + cleanup_script,
-            # Empty suite is used to prevent mmdebstrap from filtering
-            # the required packages.
-            '',
+        # Pull the ostree. Ideally the eosminbase ostree would be used to
+        # minimize the buildroot, but that's not released and would fail with
+        # --use-production-ostree.
```
wjt, line 99: A shame indeed… Maybe we could change that at some point but I don't think we want to block this change on that. (Another way to look at this would be as an incentive to make the OS as small as possible! But the builder never needs GNOME Shell.)

wjt (approved):
This looks great. One less tool to think about, one less thing to keep in sync.

Thinking about this at a high level, and assuming `R` is the OSTree ref that we are trying to build an image around:

1. Pull `R` to `/var/cache/eos-image-builder/content/ostree/repo`
2. Deploy `R` to `/var/tmp/eos-image-builder` as a mutable copy
3. Chroot into that mutable copy, bind-mounting `/var/cache/eos-image-builder/content` into the chroot
4. In `eib_ostree`, pull `R` into `/var/cache/eos-image-builder/content/ostree/repo` a second time. We hope this will be a no-op but in an edge case `R` might have been updated in the intervening seconds
5. Pull `R` locally into `/var/cache/eos-image-builder/tmp/ostree-co/ostree/repo`
6. Deploy `R` to `/var/cache/eos-image-builder/tmp/ostree-co`

It feels a bit like there is one too many pull/deploy, particularly in the case where the outermost system is also a deployment of `R`. But I think all the steps are needed!

dbnicholson:
> Thinking about this at a high level, and assuming `R` is the OSTree ref that we are trying to build an image around:
> 
> 1. Pull `R` to `/var/cache/eos-image-builder/content/ostree/repo`
> 2. Deploy `R` to `/var/tmp/eos-image-builder` as a mutable copy
> 3. Chroot into that mutable copy, bind-mounting `/var/cache/eos-image-builder/content` into the chroot
> 4. In `eib_ostree`, pull `R` into `/var/cache/eos-image-builder/content/ostree/repo` a second time. We hope this will be a no-op but in an edge case `R` might have been updated in the intervening seconds
> 5. Pull `R` locally into `/var/cache/eos-image-builder/tmp/ostree-co/ostree/repo`
> 6. Deploy `R` to `/var/cache/eos-image-builder/tmp/ostree-co`

Tis a bit much. #4 could be skipped, but it should be fast even if the remote did get updated.

lib/eib.py
dbnicholson: Totally agree. I find the multiple `for`s in comprehensions so confusing that I never use them. But then I end up doing something equally ugly like this.

wjt:
The build succeeded. Let's do it!

PR #132, opened 2023-11-29
Remove USB image support (for Endless Key V1)

wjt:
The original incarnation of Endless Key was a hybrid USB stick that contained some Kolibri stuff and a Kiwix encyclopedia usable on Windows, but also a bootable Endless OS image that could access the same content.

This is a thing of the past and its presence is confusing when we have another thing, also called Endless Key, which can be preloaded in OS images.

It has also become a practical problem because exfat-utils was removed from Debian Bookworm so image builds on that branch were broken in order to support a mode we no longer use.

Remove it all.

https://phabricator.endlessm.com/T35093


dbnicholson (approved):
Sweet

PR #133, opened 2023-11-29
Stop installing Hack apps by default

wjt:
We are no longer maintaining Hack. The apps will remain available on Flathub (unless/until they are EOLed) and installed on existing systems.

https://phabricator.endlessm.com/T35054


wjt:
I also opened https://github.com/endlessm/endless-image-config/pull/1163 that removes the Hack apps from all those configs.

dbnicholson:
> Agreed. A minor fly in the ointment is that several configurations in endless-image-config have them in their icon grids without explicitly installing the apps. That would imply that they want them to be installed. Maybe we don't care about that since basically all of those configurations are essentially dead.

Oh, I see you did that, too.

PR #134, opened 2023-11-29
buildroot: Replace exfat-utils with exfatprogs for bookworm

dbnicholson:
Now that the OS is based on bookworm, exfat-utils is no longer available and exfatprogs needs to be installed. Neither package is used directly by the image builder but rather by `eos-write-live-image`. Unfortunately, the `eos-tech-support` package it comes from does not declare the needed dependency.

https://phabricator.endlessm.com/T35035


wjt:
We don't need this script any more. #132

PR #136, opened 2023-12-05
Update gnome-initial-setup vendor configuration path

wjt:
Previously we built GNOME Initial Setup to use the path
/var/lib/eos-image-defaults/branding/gnome-initial-setup.conf and
installed the keyfile specified in the '[image] branding_fbe_config' key
there. (There was no validation that its name was actually
gnome-initial-setup.conf, which was necessary for it to work.)

When we contributed this feature upstream to gnome-initial-setup, the
path was made configurable with a build-time parameter, with no default.

Some time later, Initial Setup started reading
/etc/gnome-initial-setup/vendor.conf if present; and
/usr/share/gnome-initial-setup/vendor.conf otherwise. It retained a
build-time parameter, which we continued to set; but setting that
parameter disabled *both* of the upstream paths.

With the upstream behaviour, it is possible to ship overrides in the
immutable bit of the OS, and further override them in /etc. This is in
my opinion preferable to what we have historically done, to
always write a file in /var for all images, even if the specific image
does not want any image-specific customisation of Initial Setup.

Install any config file to the upstream path in /etc, matching a build
configuration change I will make in our gnome-initial-setup package.

Remove default gnome-initial-setup vendor.conf in favour of https://github.com/endlessm/eos-theme/pull/392.

https://phabricator.endlessm.com/T35040

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -222,7 +222,7 @@ branding_desktop_config =
 branding_desktop_logo =
 
 # Brand-specific configuration for the First Boot Experience
-branding_fbe_config = ${build:datadir}/branding/gnome-initial-setup/default/gnome-initial-setup.conf
+branding_fbe_config = 
 
 # Environment variables to substitute in branding configuration files
 branding_subst_vars_add =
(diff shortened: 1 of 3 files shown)


dbnicholson (approved):
Looks good. I'll wait on merging this until the g-i-s PRs are ready.

wjt:
I'm going to go ahead and merge this because the vanilla Debian copy of gnome-initial-setup we have in eos:master:core already uses the upstream paths.

PR #137, opened 2023-12-06
Handle updated Kolibri channels

dbnicholson:
If a Kolibri channel has been updated on Studio, the content server needs to run a channel update procedure or it won't import new content nodes.

While here, I couldn't resist the temptation to refactor all of this so that all the dispersed hooks can go away. I ran the Endless Key process locally and it seemed to work.

https://phabricator.endlessm.com/T34697


lib/eibkolibri.py
```diff
@@ -0,0 +1,134 @@
...
+    netrc_path = os.path.join(eib.SYSCONFDIR, 'netrc')
+    if not os.path.exists(netrc_path):
+        logger.info(f'No credentials in {netrc_path}')
+        return
+
+    netrc_creds = netrc(netrc_path)
+    host = urlparse(base_url).netloc
+    creds = netrc_creds.authenticators(host)
+    if not creds:
+        logger.info(f'No credentials for {host}')
```
wjt:
```python
        logger.info(f'No credentials for {host} in {netrc_path}')
```

perhaps?

lib/eibkolibri.py
```diff
@@ -0,0 +1,134 @@
...
+    with session.get(url) as resp:
+        resp.raise_for_status()
+        return resp.json()
+
+
+def wait_for_job(session, base_url, job_id):
+    """Wait for remote Kolibri job to complete"""
+    logger.debug(f'Waiting for job {job_id} to complete')
+    last_marker = None
+    while True:
```
wjt, line 41: Do we want to rate-limit this to check every 1 second rather than spamming the server as fast as we can go?

lib/eibkolibri.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # See the kolibri.core.tasks.job.State class for potential states
+        # https://github.com/learningequality/kolibri/blob/develop/kolibri/core/tasks/job.py#L17
+        status = data['status']
+        if status == 'FAILED':
+            logger.error(
+              f'Job {job_id} failed: {data["exception"]}\n{data["traceback"]}'
+            )
+            raise Exception(f'Job {job_id} failed')
+        elif status == 'CANCELED':
```
wjt: I went to check the spelling here and it really is `CANCELED` for tasks, though for Morango sync it's `CANCELLED`.

I wasn't aware of the US spelling for this word (though I guess it's consistent with other Commonwealth vs US variations) but I'm really happy to see that both spellings are used in Kolibri!

wjt (approved):
Looks great. Nice simplification too!

lib/eibkolibri.py
dbnicholson: That's fair. I might split the difference to 0.5 seconds so there's less artificial slowdown.

lib/eibkolibri.py
dbnicholson: I hadn't even thought about that! I just copied the strings directly out of the module.

PR #139, opened 2023-12-08
config: Install python3-dev package in buildroot

dbnicholson:
If the build needs to install a Python package from PyPI (like hooks/image/60-kolibri-content), it's possible that it will need to be built natively from source if there are no binary wheels available for the current Python version. That fails if the Python development environment isn't available.

https://phabricator.endlessm.com/T35125

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -84,6 +84,7 @@ packages_add =
   pigz
   python3
   python3-apt
+  python3-dev
   python3-gi
   python3-requests
   python3-venv


wjt (approved):
This seems fine, but I can never remember: why do we install packages from PyPI rather than using the Flatpak of Kolibri that we have just installed into the image?

dbnicholson:
> This seems fine, but I can never remember: why do we install packages from PyPI rather than using the Flatpak of Kolibri that we have just installed into the image?

I've never tried it myself, but last time [JP tried it](https://github.com/endlessm/eos-image-builder/issues/117#issuecomment-1716681322) it failed for him. I agree that would be a lot better.

For Endless Key it's a bit simpler since it can manage the plugins itself at runtime and all that's needed is Kolibri to drive the content import during the image build. I believe the Kolibri flatpak needs the plugins to be enabled, though.

dylanmccall:
Yeah, it would be nice to get it using Kolibri from the flatpak. But based on the two attempts to do that so far (mine, and JP's) that's probably a few days of hair pulling :b I think it's only a little more bother to fine-tune the way we preload content, such that the image has a content directory and a manifest file, instead of a content directory and a Kolibri database. Much of the complexity here comes from needing to have approximately the same Kolibri as in the flatpak, and the best thing is if ( :rainbow: some day :honey_pot: ) we stop needing to do that.

PR #140, opened 2023-12-08
Support Kolibri 0.16 tasks API

dbnicholson:
Eventually we're going to want to update our content server to Kolibri 0.16 since it will soon be the new stable version and because it has lots of bug fixes in content handling. Unfortunately, the tasks API that's used when seeding content from the image builder changed.

Initially I changed it directly to just use the 0.16 API, but then I decided I didn't want to have a flag day. This makes the image builder support both series. In addition to some manual testing, I added some unit tests to prove it out to myself. That will let us land the necessary changes in the image builder and later update the server when we feel it's appropriate (like when 0.16.0 is finally released).

This is branched from eos5.1 to ease later backporting since we'll want to have this on all active branches.

https://phabricator.endlessm.com/T34697

### DIFF
--- a/lib/eibkolibri.py
+++ b/lib/eibkolibri.py
@@ -155,25 +155,30 @@ def update_channel(session, base_url, channel_id):
 
 
 def seed_remote_channels(channel_ids):
-    """Import channels and content on remote Kolibri server"""
+    """Import channels and content on remote Kolibri server
+
+    Seeding is skipped if a custom content server is not used or there
+    are no credentials for the server. Returns True when channels were
+    seeded and False otherwise.
+    """
     config = eib.get_config()
 
     base_url = config.get('kolibri', 'central_content_base_url', fallback=None)
     if not base_url:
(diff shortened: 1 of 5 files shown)


lib/eibkolibri.py
```diff
@@ -29,6 +30,12 @@
 
 class RemoteKolibri:
     """Kolibri remote instance"""
+
+    class Series(enum.Enum):
+        """Supported Kolibri server series"""
+        KOLIBRI_0_15 = enum.auto()
+        KOLIBRI_0_16 = enum.auto()
```
wjt, line 37: A chaotic alternative would be:

```
        KOLIBRI_0_15 = 0.15
        KOLIBRI_0_16 = 0.16
```

I thought this would be rejected but actually it's completely legal and the Python documentation gives an example where enum values are tuples of floats.

https://docs.python.org/3/howto/enum.html#planet

wjt (approved):

PR #141, opened 2023-12-13
endlesskey: Import only specified content in collections

dylanmccall:
The previous behaviour, of importing all content from all channels in the collections, is preserved by setting include_full_channels to true.

https://phabricator.endlessm.com/T35103

### DIFF
--- a/config/personality/en.ini
+++ b/config/personality/en.ini
@@ -25,3 +25,13 @@ apps_add =
   com.endlessnetwork.drawingtutorials
   com.endlessnetwork.htmltutorials
   com.endlessnetwork.sciencesnacks
+
+[endlesskey]
+collections =
+  artist
+  athlete
+  curious
+  explorer
+  extras
+  inventor
+  scientist
--- a/hooks/image/53-ek-content-preload
+++ b/hooks/image/53-ek-content-preload
@@ -8,28 +8,30 @@ if [[ ! "${EIB_FLATPAK_REMOTE_FLATHUB_APPS}" =~ .*"org.endlessos.Key".* ]]; then
   exit 0
 fi
 
-# Get the list of channels to preload.
+selected_collection_files=()
+
 channels_file="${EIB_TMPDIR}"/ek-channels
 rm -f "${channels_file}"
 touch "${channels_file}"
-collections="${OSTREE_VAR}"/lib/flatpak/app/org.endlessos.Key/current/active/files/share/endless-key/collections/*.json
-for collection in ${collections}; do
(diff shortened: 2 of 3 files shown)


config/personality/en.ini
```diff
@@ -25,3 +25,12 @@ apps_add =
...
   com.endlessnetwork.sciencesnacks
+
+[endlesskey]
+collections =
+  artist
+  athlete
+  curious
+  explorer
+  inventor
+  scientist
```
dbnicholson, line 36: I think you want to also include `extras` or this won't work offline. The explore plugin will try to import all the thumbnails from nodes in the `extras` pack at runtime. See https://github.com/endlessm/kolibri-explore-plugin/blob/master/kolibri_explore_plugin/collectionviews.py#L43. I think it's fine to import whatever nodes are specified in it.

dbnicholson (approved):
This looks good modulo one comment. I'm on the fence about changing the default since it requires a synchronized change in endless-image-config, but it's not that big of a deal.

dbnicholson (changes requested):
Sorry, meant that to be request changes.

config/personality/en.ini
dylanmccall: Oh yeah, that's a good one. I pushed a change just now (75060ae1bd7b43b54b93e3dbeaf5319a126a76bb) getting it to download all thumbnails for all the included channels.

Alas, adding the `extras` collection is a bit problematic, since that manifest file includes a bunch of _content_ as well. But it probably shouldn't? I proposed https://github.com/endlessm/endless-key-collections/pull/37 just now to address that.

dbnicholson: The collections change was merged, released, and included in an org.endlessos.Key flathub release. So, I think this should be good now.

hooks/image/53-ek-content-preload
```diff
@@ -50,8 +50,14 @@ if [ -n "${EIB_KOLIBRI_CENTRAL_CONTENT_BASE_URL}" ]; then
...
 fi
 
+# Import all channel metadata and thumbnails for all channels
 for channel in $all_channels; do
   kolibri manage --skip-update importchannel network "${channel}"
+  EIB_RETRY_ATTEMPTS=2 EIB_RETRY_INTERVAL=30 eib_retry \
+    kolibri manage --skip-update \
+    importcontent --include-unrenderable-content --fail-on-error \
+    --node_ids="" --all-thumbnails \
+    network "${channel}"
```
dbnicholson, line 60: IMO, this should only be in the not full channels branch below. The full channels branch is obviously already going to import all the thumbnails. Not a big deal, though.

dbnicholson (approved):

PR #143, opened 2023-12-19
Recreate /var

wjt:
Sadly https://github.com/endlessm/eos-ostree-builder/pull/189 broke the image builder, because we create the buildroot by deploying an ostree, and the various things that set up `/var` on a fresh system are not all done here.

https://phabricator.endlessm.com/T4482

### DIFF
--- a/run-build
+++ b/run-build
@@ -189,6 +189,9 @@ class ImageBuildRoot(object):
             os.path.join(self.builddir, 'var/lib/dpkg'),
         )
 
+        # update-catalog requires /var/lib/sgml-base to exist
+        os.makedirs(os.path.join(self.builddir, 'var/lib/sgml-base'))
+
         log.info('Installing buildroot packages')
         apt_env = os.environ.copy()
         apt_env['DEBIAN_FRONTEND'] = 'noninteractive'
--- a/run-build
+++ b/run-build
@@ -42,7 +42,7 @@ log = logging.getLogger(os.path.basename(__file__))
 class ImageBuildRoot(object):
     """Build root for performing actual image build
 
-    A temporary build root is created with mmdebstrap at builddir. The
+    A temporary build root is created with ostree at builddir. The
     options in the config section buildroot control what gets installed
     and mounted in the buildroot.
     """
--- a/run-build
+++ b/run-build
@@ -173,6 +173,22 @@ class ImageBuildRoot(object):
         for path in to_mount:
             self.mount(path)
 
+        # Remove the tmpfiles.d snippet that sets up symlinks for changes we have
+        # just reversed, and for the dpkg database we are about to move back
+        os.unlink(os.path.join(self.builddir, 'usr/lib/tmpfiles.d/ostree.conf'))
+
+        # Create any files that would normally be handled at runtime. This is
+        # particularly important for /var since it's likely that it was removed
+        # from the ostree commit. Run this inside the chroot to avoid depending on
+        # systemd-tmpfiles on the host system.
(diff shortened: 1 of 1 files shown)


wjt:
This is a draft because the build now fails later on while running `update-catalog`, triggered by some package installing an SGML catalog I guess:

```
Processing triggers for sgml-base (1.31) ...
cannot open /var/lib/sgml-base/supercatalog.new for writing: No such file or directory at /usr/sbin/update-catalog line 313.
dpkg: error processing package sgml-base (--configure):
 installed sgml-base package post-installation script subprocess returned error exit status 2
Errors were encountered while processing:
 docutils-common
 python3-docutils
 awscli
 sgml-base
```

wjt:
Alright after creating a load of random files I now get this far:

```
+ 10:45:25 eib_image: ostree --repo=/var/cache/eos-image-builder/tmp/ostree-co/ostree/repo pull-local --disable-fsync --remote=eos /var/cache/eos-image-builder/content/ostree/eos os/eos/amd64/master
9079 metadata, 100452 content objects imported; 0 bytes content written
+ 10:45:31 eib_image: ostree --repo=/var/cache/eos-image-builder/tmp/ostree-co/ostree/repo pull eos ostree-metadata
error: open(O_TMPFILE): No such file or directory
```

wjt:
This is effectively the same change as https://github.com/endlessm/eos-container-builder/pull/26

run-build
```diff
@@ -173,6 +173,28 @@ class ImageBuildRoot(object):
...
+        os.rename(
+            os.path.join(self.builddir, 'usr/share/dpkg/database'),
+            os.path.join(self.builddir, 'var/lib/dpkg'),
+        )
+
+        # update-catalog requires /var/lib/sgml-base to exist
+        os.makedirs(os.path.join(self.builddir, 'var/lib/sgml-base'))
+
+        # update-xmlcatalog requires /var/lib/xml-core to exist
+        os.makedirs(os.path.join(self.builddir, 'var/lib/xml-core'))
```
wjt, line 196: The comment in `run-build` seems prescient:

```
        # Pull the ostree. Ideally the eosminbase ostree would be used to
        # minimize the buildroot, but that's not released and would fail with
        # --use-production-ostree.
```

I haven't checked but I would be interested to know if these packages are part of eosminbase. If not then these steps would not be necessary.

wjt: Actually I have a vague memory of seeing errors from `sgml-base` in my toolboxes.

dbnicholson: Are the errors fatal? Ideally those would have tmpfiles snippets so we didn't need to do anything like this. Or the scripts would, you know, attempt to create the directories instead of assuming they exist? But meh, this is fine.

dbnicholson (approved):
Thanks for this. I did not anticipate this fallout even though I should have since I worked through it in the container builder and this is basically the same thing.

A future improvement would be to actually use the containers instead of bespoke `chroot` setups. Then the container builder would be the one place where you'd do the conversion from immutable OS to mutable OS.

dbnicholson (approved):

run-build
wjt: Yeah, unfortunately they are fatal, and a trigger failing is in turn fatal to the `apt-get install` invocation.

I agree that ideally they would have tmpfiles snippets – though I think the Way of the Future in OSTree is to put it in `/usr/share/factory` or whatever it's called?

wjt:
> A future improvement would be to actually use the containers instead of bespoke `chroot` setups. Then the container builder would be the one place where you'd do the conversion from immutable OS to mutable OS.

That would indeed be good!

PR #149, opened 2024-01-18
Remove Cheese from mandatory apps

wjt:
Cheese was moved from the OS to a Flatpak in eos4.0. There has been a checkpoint between eos4.0 and eos5.0+, so we can now remove the flatpak-autoinstall.d rule that requires it to be installed. It follows that we can relegate it to "default" status, which in particular means that base images need no longer include it.

Depends on:

- [ ] https://github.com/endlessm/eos-application-tools/pull/83

https://phabricator.endlessm.com/T35175

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -475,7 +475,6 @@ repo_file = https://dl.flathub.org/repo/flathub.flatpakrepo
 apps_add_mandatory =
   org.chromium.Chromium
   org.gnome.Calculator
-  org.gnome.Cheese
   org.gnome.Contacts
   org.gnome.Epiphany
   org.gnome.FileRoller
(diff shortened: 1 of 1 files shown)


manuq (approved):
I see, so "apps_add_mandatory" apps are going to the base image, and "apps_add" go to default images like GIMP, Godot, etc.

wjt:
> I see, so "apps_add_mandatory" apps are going to the base image, and "apps_add" go to default images like GIMP, Godot, etc.

The way it works is that `base.ini` uses the `${apps_add_defaults}` list in its `apps_del` key for each remote.

https://github.com/endlessm/eos-image-builder/blob/master/config/personality/base.ini#L8-L14

The image builder adds up all the `apps_add*` keys for a remote, then subtracts all the `apps_del*` keys for that remote.

PR #156, opened 2024-03-05
Fix dpkg queries

dbnicholson:
The ostree commits now ship with an empty `/var`, which means that `/var/lib/dpkg/status` doesn't exist. Use `/usr/share/dpkg/database` instead, which is what `/var/lib/dpkg` is a symlink to at runtime.

https://phabricator.endlessm.com/T35244


wjt (approved):
Very nice!

PR #158, opened 2024-03-26
defaults: add Decibels & Music

cassidyjames:
https://phabricator.endlessm.com/T32696

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -466,6 +466,7 @@ apps_add_mandatory =
 [flatpak-remote-flathub]
 repo_file = https://dl.flathub.org/repo/flathub.flatpakrepo
 apps_add_mandatory =
+  com.vixalien.decibels
   org.chromium.Chromium
   org.gnome.Calculator
   org.gnome.Contacts
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -473,7 +473,6 @@ apps_add_mandatory =
   org.gnome.Epiphany
   org.gnome.FileRoller
   org.gnome.Loupe
-  org.gnome.Music
   org.gnome.Totem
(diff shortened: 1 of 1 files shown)


wjt (changes requested):
I don't think these should be mandatory, I think they should be in `apps_add` (i.e. not present in the `base` image), as Rhythmbox is.

And, presumably you intended to remove Rhythmbox from the defaults?

cassidyjames:
@wjt ah, I saw Totem there and thought Decibels served a similar purpose; without it, I guess audio files will just open in Totem? I agree Music should replace Rhythmbox in `apps_add`, though; my mistake.

config/defaults.ini
```diff
@@ -502,7 +503,7 @@ apps_add =
   org.gimp.GIMP
   org.gnome.Cheese
   org.gnome.Gnote
-  org.gnome.Rhythmbox3
+  org.gnome.Music
```
wjt, line 506: Very convenient that they fall in the same place in the sort order!

config/defaults.ini
```diff
@@ -466,6 +466,7 @@ apps_add_mandatory =
 [flatpak-remote-flathub]
 repo_file = https://dl.flathub.org/repo/flathub.flatpakrepo
 apps_add_mandatory =
+  com.vixalien.decibels
```
wjt, line 469: Yes, on second thoughts, your line of reasoning makes sense: being able to play audio files is essential functionality, so decibels should go here; while having a music library app is not.

wjt (approved):

config/defaults.ini
cassidyjames: Indeed!

PR #162, opened 2024-04-20
Support QEMU qcow2 image format

dbnicholson:
It's a nice format for loading in GNOME Boxes or any VM tool that uses QEMU. In particular, it's the native format used by GNOME Boxes.

https://phabricator.endlessm.com/T35332

### DIFF
--- a/stages/eib_image
+++ b/stages/eib_image
@@ -597,6 +597,27 @@ EOF
     "${EIB_HELPERSDIR}"/create-vm-image "$(eib_outfile ovf.zip)" "${img}"
   fi
 
+  # Create qcow2 image if required.
+  if [ "${EIB_IMAGE_QCOW2}" = true ]; then
+    local qcow2
+
+    # Convert the raw image to qcow2 with compression (-c). This uses
+    # zlib, but it would be better to use zstd with -o
+    # compression_type=zstd. zstd support was added as an option in QEMU
+    # 5.1 in 2020, but we'd want to make sure zstd support was included
+    # in any distros/flatpaks we cared about.
+    qcow2=$(eib_outfile qcow2)
+    qemu-img convert -c -f raw -O qcow2 "${img}" "${qcow2}"
+
+    # Since the format is copy-on-write, we can grow it as much as we
+    # want without increasing the download size. Add 10 GiB.
+    qemu-img resize -f qcow2 "${qcow2}" +10G
+
+    stat -c "%s" "${qcow2}" > "${qcow2}.size"
+    sign_file "${qcow2}" &
+    checksum_file "${qcow2}" &
+  fi
+
   # Await signing and checksumming subprocesses
   wait
 
--- a/config/arch/amd64.ini
+++ b/config/arch/amd64.ini
@@ -10,6 +10,5 @@ packages_add =
   grub2
   mtools
   squashfs-tools
-  qemu-utils
   xorriso
   zip
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -88,6 +88,7 @@ packages_add =
   python3-gi
   python3-requests
   python3-venv
+  qemu-utils
   rsync
   udev
   util-linux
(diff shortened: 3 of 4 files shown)


stages/eib_image
```diff
@@ -597,6 +597,22 @@ EOF
...
+  # Create qcow2 image if required.
+  if [ "${EIB_IMAGE_QCOW2}" = true ]; then
+    local qcow2
+
+    qcow2=$(eib_outfile qcow2)
+    qemu-img convert -c -f raw -O qcow2 "${img}" "${qcow2}"
+
+    # Since the format is copy-on-write, we can grow it as much as we
+    # want without increasing the download size. Add 10 GiB.
+    qemu-img resize -f qcow2 "${qcow2}" +10G
```
wjt, line 609: OK this addresses my first reflexive reaction: whenever I convert an image to qcow2 manually (which I do relatively frequently!) I forget to do this and it's really annoying to go back and do it later.

stages/eib_image
```diff
@@ -597,6 +597,22 @@ EOF
     "${EIB_HELPERSDIR}"/create-vm-image "$(eib_outfile ovf.zip)" "${img}"
   fi
 
+  # Create qcow2 image if required.
+  if [ "${EIB_IMAGE_QCOW2}" = true ]; then
+    local qcow2
+
+    qcow2=$(eib_outfile qcow2)
+    qemu-img convert -c -f raw -O qcow2 "${img}" "${qcow2}"
```
wjt, line 605: I recently learned that qcow2 natively supports zstd compression. No idea how good it is but:

```suggestion
    qemu-img convert -c -f raw -O qcow2 -o compression_type=zstd "${img}" "${qcow2}"
```

wjt (commented):
This would be useful, but we already build & store 3 massive files for each `eos` build – it slows down the build process every night and occupies bytes on a disk somewhere in the sky forever.

I have a hunch that the VMDK images are essentially unused. I don't think we can determine from OS metrics/pings/activations whether a VM was installed from the VMDK or from an img/ISO. But maybe we can look at our CDN logs and see if anyone has ever downloaded them…

stages/eib_image
dbnicholson: How recently has QEMU/qcow2 supported this? I'd be worried about trying to use the image on an older host that doesn't have zstd support.

dbnicholson:
> This would be useful, but we already build & store 3 massive files for each `eos` build – it slows down the build process every night and occupies bytes on a disk somewhere in the sky forever.
> 
> I have a hunch that the VMDK images are essentially unused. I don't think we can determine from OS metrics/pings/activations whether a VM was installed from the VMDK or from an img/ISO. But maybe we can look at our CDN logs and see if anyone has ever downloaded them…

Fair point. I can try to dig some download logs today and see if they're ever downloaded. Personally I've never used Virtualbox, but almost certainly the OVF images are more useful on Windows or OSX.

Alternatively, I could limit it to `base` with the thought that this is primarily a developer tool at the moment.

stages/eib_image
dbnicholson: One thing I've always wanted is a simpler way to trigger the initramfs partitioner to run again to support this kind of after the fact VM resizing.

stages/eib_image
wjt: Since 5.1 https://wiki.qemu.org/ChangeLog/5.1 released in August 2020 https://www.qemu.org/2020/08/11/qemu-5-1-0/.

OK, now I understand that the `-c` above will compress with `zlib`. I incorrectly believed that `-o compression_type=...` is what enabled compression. So I think `zlib` is fine.

dsd:
I think limiting to base makes sense.

stages/eib_image
dbnicholson: If you're "in the know", you can add the [marker attribute](https://github.com/endlessm/eos-boot-helper/blob/master/dracut/repartition/endless-repartition.sh#L99) back to the partition so it's resized on the next boot. You also need to delete [`/var/eos-booted`](https://github.com/endlessm/eos-boot-helper/blob/master/eos-firstboot.service#L8) or manually run `resize2fs` after that. But it would be a lot more convenient if there was a tool do that.

dbnicholson:
> I think limiting to base makes sense.

I'll update the PR to do that and we can punt on OVF images.

stages/eib_image
dbnicholson: Indeed. I'll add a comment about exactly what the convert command is doing.

dbnicholson:
PR updated with moar comments + limiting to EOS base.

wjt (approved):
moo

PR #164, opened 2024-05-29
image: Add hook to customize metrics server URLs

dbnicholson:
Allow configuring the metrics server URLs so that 3rd parties deploy the Endless metrics stack they can build images to use them out of the box.

https://phabricator.endlessm.com/T35420


dbnicholson:
I'm marking this as a draft since we may not want to bother with the eos-phone-home dropin if https://github.com/endlessm/eos-phone-home/pull/18 is merged.

cassidyjames:
https://github.com/endlessm/eos-phone-home/pull/18 indeed seems cleaner, so I think I prefer that unless there's a good reason to do it this way

hooks/image/50-metrics-urls
```diff
@@ -0,0 +1,19 @@
...
+  dropin_dir="${OSTREE_DEPLOYMENT}"/etc/systemd/system/eos-phone-home.service.d
+  dropin_file="${dropin_dir}"/host.conf
+  mkdir -p "${dropin_dir}"
+  cat > "${dropin_file}" <<EOF
+# Override the default eos-phone-home server URL.
+[Service]
+ExecStart=
+ExecStart=/usr/lib/eos-phone-home/eos-phone-home --host=${EIB_IMAGE_ACTIVATION_SERVER_URL}
+EOF
+fi
```
wjt, line 19: We can now revise this to write a config file.

dbnicholson:
I hate the idea of editing an INI file with `sed` and `echo` so I had a brief foray into "proper" editing using `configparser`. But it turns out that it's [write](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.write) method is fairly barebones, so I scrapped it. This should be good enough knowing that both files only have a single `global` section.

wjt (approved):

PR #167, opened 2024-08-21
Don't install BIOS GRUB in sd-boot images

wjt:
systemd-boot is EFI-only. If we use systemd-boot on an image, we
actively do not want the system to be BIOS-bootable. (And the installed
BIOS GRUB doesn't actually work on these images anyway.)

On sd-boot images, don't install any BIOS bootloader and don't create a BIOS Boot partition.

In addition, drop code for the long-unused `cros` partition layout.

https://phabricator.endlessm.com/T34775

### DIFF
--- a/stages/eib_image
+++ b/stages/eib_image
@@ -209,8 +209,10 @@ create_image() {
       i386|amd64)
         # EFI system partition
         echo "size=${esp_size}MiB, type=C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
-        # GRUB BIOS BOOT partition
-        echo "size=1MiB, type=21686148-6449-6E6F-744E-656564454649"
+        if [[ "${EIB_IMAGE_SDBOOT}" != "true" ]]; then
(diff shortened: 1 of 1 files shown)


wjt:
Scratch build running at https://ci.endlessos.org/job/image-build-amd64/30438

wjt:
Tested that image in 2 ways:

1. Booted qcow2 image in Boxes. It works correctly and the root filesystem is correctly resized to fill the disk.
2. Installed the `.img.xz` on a sacrificial laptop using eosinstaller. eos-installer correctly recognised that it is an Endless OS disk even though the BIOS Boot partition is missing. This is expected because we dropped that requirement in 2020 for GNOME OS – the installer now only checks that the disk has at least 2 partitions: ESP and OS. The resulting system booted correctly and the root filesystem was correctly expanded to fill the disk.

dbnicholson (approved):
Makes sense, but is obviously instills a bit of fear. Godspeed!

PR #168, opened 2024-11-19
Hooks: Add hook to expose Endless Key launchers in app grid

cassidyjames:
Is this the right idea? It sort of feels dirty to include a whole shell script here. And do I need to make it executable?

https://phabricator.endlessm.com/T35614

### DIFF
--- a/hooks/image/54-ek-launchers
+++ b/hooks/image/54-ek-launchers
@@ -1,0 +1,10 @@
+# Expose Endless Key launchers in the app grid (once a specific content pack has been downloaded)
+
+mkdir -p "${OSTREE_DEPLOYMENT}/etc/systemd/user-environment-generators"
+cat <<- 'EOF' > "${OSTREE_DEPLOYMENT}/etc/systemd/user-environment-generators/62-endless-key.sh"
+	#!/bin/bash
+
+	XDG_DATA_DIRS="/var/lib/endless-key/data/content/xdg/share:${XDG_DATA_DIRS:-/usr/local/share:/usr/share}"
+	echo "XDG_DATA_DIRS=$XDG_DATA_DIRS"
+EOF
+chmod 755 "${OSTREE_DEPLOYMENT}/etc/systemd/user-environment-generators/62-endless-key.sh"


hooks/image/54-ek-launchers
```diff
@@ -0,0 +1,12 @@
+# Expose Endless Key launchers in the app grid (once a specific content pack has been downloaded)
+
+if [[ ! "${EIB_FLATPAK_REMOTE_FLATHUB_APPS}" =~ .*"org.endlessos.Key".* ]]; then
+  exit 0
+fi
+
+cat <<- EOF > /etc/systemd/user-environment-generators/62-endless-key.sh
```
wjt, line 7: This hook is not running inside the image, but outside it. 
```suggestion
cat <<- EOF > "${OSTREE_DEPLOYMENT}/etc/systemd/user-environment-generators/62-endless-key.sh"
```

hooks/image/54-ek-launchers
```diff
@@ -0,0 +1,12 @@
...
+  #!/bin/bash
```
wjt, line 8: Does `<<-` strip all leading whitespace, or only leading tabs?

hooks/image/54-ek-launchers
wjt: I think you can avoid needing to backslash-escape the `$`s by single quoting the delimiter, i.e.

```suggestion
cat <<- 'EOF' > "${OSTREE_DEPLOYMENT}/etc/systemd/user-environment-generators/62-endless-key.sh"/etc/systemd/user-environment-generators/62-endless-key.sh"
```

hooks/image/54-ek-launchers
```diff
@@ -0,0 +1,12 @@
+# Expose Endless Key launchers in the app grid (once a specific content pack has been downloaded)
+
+if [[ ! "${EIB_FLATPAK_REMOTE_FLATHUB_APPS}" =~ .*"org.endlessos.Key".* ]]; then
```
wjt, line 3: This is nice but I think not necessary, it's not an error for a path in XDG_DATA_DIRS not to exist.

wjt:
It doesn't need to be executable. If you look at the other files in this directory, only the ones ending in `.chroot` are executable. That's because they are run in a different environment, by chrooting into the target image, and then invoking them as a program. The non-chroot ones are sourced with the `.` operator.

hooks/image/54-ek-launchers
cassidyjames: Ah, the docs say tabs. Good catch. Yay invisible deterministic whitespace!

wjt:
I agree that it's a little ugly to write a shell script to  write an embedded shell script to a place, but for better or for worse that's how it's done for now.

You could imagine instead putting the file in the OS, and then just making a symlink here. But I think this is OK.

hooks/image/54-ek-launchers
wjt: (The conditionality is that only the image where this is wanted will list this hook.)

hooks/image/54-ek-launchers
```diff
@@ -0,0 +1,8 @@
+# Expose Endless Key launchers in the app grid (once a specific content pack has been downloaded)
+
+cat <<- 'EOF' > "${OSTREE_DEPLOYMENT}/etc/systemd/user-environment-generators/62-endless-key.sh"
+	#!/bin/bash
+
+	XDG_DATA_DIRS="/var/lib/endless-key/data/content/xdg/share:${XDG_DATA_DIRS:-/usr/local/share:/usr/share}"
+	echo "XDG_DATA_DIRS=$XDG_DATA_DIRS"
+EOF
```
dbnicholson, line 8: I think the generated script needs to be made executable or systemd won't actually execute it at runtime.
```suggestion
EOF
chmod 755 "${OSTREE_DEPLOYMENT}/etc/systemd/user-environment-generators/62-endless-key.sh"
```

dbnicholson (commented):
This feels like the kind of thing that should be in the OS. In fact, we already have this exact thing for [regular Kolibri](https://github.com/endlessm/eos-kolibri/blob/master/data/user-environment-generators/61-eos-kolibri.sh.in).

cassidyjames:
I threw a test build at Jenkins, and it errored out with:

```
++ 21:07:53 54-ek-launchers: cat
/tmp/image-build-PJzO8VoBgE/eos-image-builder/hooks/image/54-ek-launchers: line 3: /var/cache/eos-image-builder/tmp/ostree-co/ostree/deploy/eos/deploy/18fedf9dcbf070e57e12ff5a93621573349ee8276dbc9717629fbb52f295df56.0/etc/systemd/user-environment-generators/62-endless-key.sh: No such file or directory
```

dylanmccall:
> This feels like the kind of thing that should be in the OS. In fact, we already have this exact thing for [regular Kolibri](https://github.com/endlessm/eos-kolibri/blob/master/data/user-environment-generators/61-eos-kolibri.sh.in).

Indeed, another approach here is we change eos-kolibri to also build its own `61-eos-endless-key.sh` environment generator right there, but in that one we add a test at the beginning, maybe for a file like `/var/lib/endless-key/data/content/xdg/show-launchers`. The image builder hook would just need to add that file. Less code that we can't update :)

I'm not too bothered by this implementation, but I'm curious if this is going to be needed for a lot of images, or if it's just a one-off?

hooks/image/54-ek-launchers
```diff
@@ -0,0 +1,8 @@
+# Expose Endless Key launchers in the app grid (once a specific content pack has been downloaded)
+
+cat <<- 'EOF' > "${OSTREE_DEPLOYMENT}/etc/systemd/user-environment-generators/62-endless-key.sh"
```
wjt, line 3: The test build failed because this directory doesn't exist. 

```suggestion
mkdir -p "${OSTREE_DEPLOYMENT}/etc/systemd/user-environment-generators"
cat <<- 'EOF' > "${OSTREE_DEPLOYMENT}/etc/systemd/user-environment-generators/62-endless-key.sh"
```

wjt (changes requested):

PR #169, opened 2024-11-21
Add config inheritance for personality field

wjt:
It is often the case that we want to build an image which just applies a
few minor changes to an existing personality.  For instance, the apps we
want for most Spanish images are the same, but we have a few
country-specific applications, and we often wish to build a sub-variant
for a partner or region of a particular country.

Currently we need to copy the full list of apps & settings from our
es.ini personality config into (for instance) es_MX.ini; and then again
from es_MX.ini into (e.g.) es_MX_aguascalientes.ini to add some
region-specific apps or resources

Implement a form of inheritance where (e.g.) building the es_MX
personality reads es.ini and es_MX.ini, in that order; and (e.g.)
building the es_MX_aguascalientes personality reads es.ini, es_MX.ini
and es_MX_aguascalientes.ini, again in that order.  This matches the
convention that we have of naming personalities for the locale, and then
adding further underscore-separated suffixes for subvariants thereof.

This is only implemented for the personality field. This is where we do
most of our customisation and is the place where inheritance would be
most useful.

https://phabricator.endlessm.com/T34731

### DIFF
--- a/run-build
+++ b/run-build
@@ -387,6 +387,27 @@ class ImageBuilder(object):
             datetime.datetime.utcnow().strftime('%y%m%d-%H%M%S')
         )
 
+    @staticmethod
+    def _get_prefixes(value):
+        """Yields all prefixes of value split by _, including the whole string.
+
+        >>> list(ImageBuilder._get_prefixes("en_GB_orkney"))
+        ["en", "en_GB", "en_GB_orkney"]
+        """
+        parts = value.split("_")
+        for i in range(1, len(parts) + 1):
+            yield "_".join(parts[:i])
+
+    def _get_attr_values(self, attr):
+        """Yields one or more values for attr. For most properties this is a single
+        value, but for "personality" this yields a series of values to implement
+        inheritance."""
+        value = getattr(self, attr)
+        if attr == "personality":
+            yield from self._get_prefixes(value)
+        else:
+            yield value
+
     def _get_config_paths(self, dir_path, dir_ns):
         config_files = []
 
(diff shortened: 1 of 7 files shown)


run-build
```diff
@@ -387,6 +387,27 @@ class ImageBuilder(object):
...
+        for i in range(1, len(parts) + 1):
+            yield "_".join(parts[:i])
+
+    def _get_attr_values(self, attr):
+        """Yields one or more values for attr. For most properties this is a single
+        value, but for "personality" this yields a series of values to implement
+        inheritance."""
+        value = getattr(self, attr)
+        if attr == "personality":
+            yield from self._get_prefixes(value)
```
dbnicholson, line 407: I kinda think we should do this for all fields. Certainly for product and platform it could make sense to have a specialized version that tweaks the normal version without having to abuse personality. For branch and arch it probably doesn't make sense. Just starting with personality would be fine, though.

dbnicholson:
I'm all for this and make a quick attempt at this a while back that I never finished.

run-build
wjt: I originally wrote it as working for all fields. We can obviously do that but I found it easier to write the documentation & explain what the feature is for if it was just the personality field.

wjt:
It turns out the hard parts are:

- going through all the configs and updating them. Because this inheritance is implicit rather than explicit it's a flag day and everything has to be checked & updated. I'm nearly there, but...
- making the list of kolibri channel IDs a merged list causes it to be sorted which I believe will actually affect the order the channels are presented in Kolibri in the image. Personally I don't care but I have some memory that it was deliberate and someone stopped me from sorting the channel IDs in the config for this reason.

dbnicholson:
> It turns out the hard parts are:
> 
> * going through all the configs and updating them. Because this inheritance is implicit rather than explicit it's a flag day and everything has to be checked & updated. I'm nearly there, but...

Yeah, that is probably very annoying. I don't envy you...

> * making the list of kolibri channel IDs a merged list causes it to be sorted which I believe will actually affect the order the channels are presented in Kolibri in the image. Personally I don't care but I have some memory that it was deliberate and someone stopped me from sorting the channel IDs in the config for this reason.

It is [deliberate](https://github.com/endlessm/eos-image-builder/blob/master/hooks/image/60-kolibri-content#L80). Hmm, I think I have a solution for the sorted merge...

dbnicholson:
> It is [deliberate](https://github.com/endlessm/eos-image-builder/blob/master/hooks/image/60-kolibri-content#L80). Hmm, I think I have a solution for the sorted merge...

See #170.

run-build
dbnicholson: Makes sense. We can always expand if it seems useful.

run-build
```diff
@@ -398,20 +419,20 @@ class ImageBuilder(object):
...
-            namespace = (dir_ns +
-                         '_'.join((attr1, attr2, getattr(self, attr1),
-                                   getattr(self, attr2))))
-            config_files.append((path, namespace))
+            for (value1, value2) in itertools.product(self._get_attr_values(attr1),
+                                                      self._get_attr_values(attr2)):
+                ini_name = (value1 + '-' + value2 + '.ini')
+                path = os.path.join(dir_path, attr1 + '-' + attr2, ini_name)
+                namespace = (dir_ns +
+                             '_'.join((attr1, attr2, value1, value2)))
```
dbnicholson, line 434: Ugh, string joins with `+`. Older Dan is ashamed of younger Dan.

dbnicholson (approved):
Looks like a winner!

PR #170, opened 2024-11-22
ImageConfigParser: Don't sort merged values

dbnicholson:
The order of configuration values may be significant, so the parser should not sort them. To do that, use a single counter for add and del variants and emit the values that have positive counts without sorting. This relies on the Python dict stable order, which has been the case since Python 3.6.

I've done no real world testing of this. It should also be stated that the order you get is "first merged section wins", so it's not really reliable. For example:

```
[section]
# From product config
option_add_product = foo
# From local product config
option_add_local_product = bar
```

You'd get `foo\nbar` even though you might really prefer `bar\nfoo` from your local setting. With explicit sorting you'd always get the same values regardless of what order the unmerged options came in.

### DIFF
--- a/lib/eib.py
+++ b/lib/eib.py
@@ -242,25 +242,23 @@ class ImageConfigParser(configparser.ConfigParser):
                                  opt)
                     yield (section, opt)
             else:
-                add_vals = Counter()
+                values = Counter()
                 for opt in add_opts:
                     logger.debug('Adding %s %s values from %s', section,
                                  option, opt)
-                    add_vals.update(sect[opt].split())
+                    values.update(sect[opt].split())
                     yield (section, opt)
 
-                del_vals = Counter()
                 for opt in del_opts:
                     logger.debug('Removing %s %s values from %s', section,
                                  option, opt)
-                    del_vals.update(sect[opt].split())
+                    values.subtract(sect[opt].split())
                     yield (section, opt)
 
-                # Set the option to the difference of the counters.
+                # Set the option to the keys that have positive values.
                 # Merge the values together with newlines like they were
                 # in the original configuration.
-                vals = add_vals - del_vals
-                sect[option] = '\n'.join(sorted(vals.keys()))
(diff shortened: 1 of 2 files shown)


wjt:
I'll look more closely at this on Monday.

At a first glance it looks very neat, and it solves part of the problem discussed on #169 – without this change, if we make the Kolibri channels field mergeable then it's impossible to preserve order.

The harder part is that in the derived-personality case, you might want (e.g.) `en_GB.ini` to add 2 more channels to the set provided by `en.ini`, but you may not want them to go after all the channels from `en.ini`. I think the way you would deal with that is to first `_del` all the inherited channels (removing them from the counter), then explicitly add them all back along with the 2 extra ones. Or explicitly set the field (which I think overrides the merging? or could be made to). Not ideal but it's not (substantially) worse than what we have now.

wjt:
I rebased this backwards so that it can be merged onto `eos6.0`.

lib/eib.py
```diff
@@ -242,25 +242,23 @@ def _merge_option(self, section_pattern, option):
...
+                    values.subtract(sect[opt].split())
                     yield (section, opt)
 
-                # Set the option to the difference of the counters.
+                # Set the option to the keys that have positive values.
                 # Merge the values together with newlines like they were
                 # in the original configuration.
-                vals = add_vals - del_vals
-                sect[option] = '\n'.join(sorted(vals.keys()))
+                sect[option] = '\n'.join(k for k, v in values.items() if v > 0)
```
wjt, line 261: I was curious whether `Counter` provides a shorter way to spell “only keys that have positive values” and it does:


```python
In [7]: values = Counter(foo=3, bar=-1)

In [8]: +values
Out[8]: Counter({'foo': 3})

In [9]: '\n'.join(+values)
Out[9]: 'foo'
```

So we could write:

```suggestion
                sect[option] = '\n'.join(+values)
```

But I've never seen this before and I think it's too obscure.

wjt (approved):

PR #171, opened 2025-01-11
Match selected Endless Key content collections correctly

wjt:
EIB_ENDLESSKEY_COLLECTIONS is a whitespace (typically space or newline)-separated list of collection names. These are matched against collection filenames to determine what to fetch.

We have collection files named `extras-preload-0001.json` and `spanish-extras-preload-0001.json`. The intention is that if `[endlesskey] collections` contains `spanish-extras-preload` but not `extras-preload`, only the latter will be imported, not the former.

However, the logic that matched filenames against the configured array did not consider the case when one collection name is a substring of another. It would incorrectly match `extras-preload-0001.json` against `spanish-extras-preload` because the latter contains `extras-preload` as a substring.

Replace runs of whitespace in EIB_ENDLESSKEY_COLLECTIONS with colons, with an extra colon at the start and end of the resulting string. Then, surround the base collection name from each collection manifest path with colons in the same way, and check if this is a substring. This assumes that the colon does not appear in collection names: a rather safe assumption I think.

(The root cause of this bug is that shell scripts are not a suitable language for non-trivial logic. I tried to use Bash arrays but apparently Bash does not have a "contains" operator for arrays. I also considered using the ASCII unit separator character, U+001F, as the delimiter, which is even less likely to appear in collection names than a colon, but decided against it.)

https://phabricator.endlessm.com/T35668#1004203

(PR targeting eos6.0 so that I can test it. I'll rebase it to the closest common ancestor of eos6.0 and master to be merged directly into both if it works.)

### DIFF
--- a/hooks/image/53-ek-content-preload
+++ b/hooks/image/53-ek-content-preload
@@ -17,8 +17,8 @@ all_collection_files="${OSTREE_VAR}"/lib/flatpak/app/org.endlessos.Key/current/a
 for collection_file in ${all_collection_files}; do
   # Check if the file basename stripped off of -0001.json is part of the list
   # of collections to be installed.
-  bn=$(basename ${collection_file})
-  if [[ "${EIB_ENDLESSKEY_COLLECTIONS}" =~ .*"${bn%-????.json}".* ]] ; then
+  bn=$(basename "${collection_file%-????.json}")
+  if [[ " ${EIB_ENDLESSKEY_COLLECTIONS} "  =~ [[:space:]]${bn}[[:space:]] ]] ; then
     selected_collection_files+=("${collection_file}")
     jq -r '.channels[].id' "${collection_file}" >> "${channels_file}"
   fi


hooks/image/53-ek-content-preload
```diff
@@ -14,11 +14,13 @@ channels_file="${EIB_TMPDIR}"/ek-channels
...
 all_collection_files="${OSTREE_VAR}"/lib/flatpak/app/org.endlessos.Key/current/active/files/share/endless-key/collections/*.json
+# Convert whitespace-separated list of collections to a colon-delimited string
+requested_collections=":$(echo -n "${EIB_ENDLESSKEY_COLLECTIONS}" | tr -s '[:space:]' ':'):"
 for collection_file in ${all_collection_files}; do
   # Check if the file basename stripped off of -0001.json is part of the list
   # of collections to be installed.
-  bn=$(basename ${collection_file})
-  if [[ "${EIB_ENDLESSKEY_COLLECTIONS}" =~ .*"${bn%-????.json}".* ]] ; then
+  bn=$(basename "${collection_file%-????.json}")
+  if [[ ${requested_collections} =~ :${bn}: ]] ; then
```
dbnicholson, line 23: Switching to `:` delimited is reasonable. However, here's an alternative that I believe does the right thing:

```
[[ ${EIB_ENDLESSKEY_COLLECTIONS} =~ (^|[[:space:]])${bn}($|[[:space:]]) ]]
```
or
```
[[ " ${EIB_ENDLESSKEY_COLLECTIONS} " =~ [[:space:]]${bn}[[:space:]] ]]
```

These would be simpler with the `\b` boundary class available in `grep` and perl REs.

wjt:
Ah, that's nicer than messing with `tr`. Good thinking! (I guess using the
same character class with tr should have tipped me off.)

I did try using boundary matches first but found that none of the syntaxes
I know of work. (Actually, does \b not consider a hyphen a word boundary in
those other tools?)

wjt:
Reworked this to use `[[:space:]]`. The test build yesterday (with the colon-based implementation) worked.

dbnicholson:
> (Actually, does \b not consider a hyphen a word boundary in those other tools?)

Heh, you're probably right! I mostly use it in one-off queries and ignore the false positives.

dbnicholson (approved):
That looks correct to me.

wjt:
Also merged to master as cc21f84e.

PR #172, opened 2025-01-17
Populate .desktop files for Endless Key channels

wjt:
Endless Key was designed to be used via a single entrypoint, rather than per-channel launchers as we have for Kolibri on Endless OS. However, a partner have specifically requested per-channel launchers coupled with the Endless Key interface & content. To support them, Endless Key bundles the same xdg plugin for Kolibri as for the kolibri-gnome app, which generates .desktop files etc. into `/var/lib/endless-key/data/content/xdg/share/`. Then in this partner's OS image, we drop a file into /etc/systemd/user-environment-generators to add this directory to XDG_DATA_DIRS, just like the file we ship in /usr/systemd/user-environment-generators for Kolibri.

However, previously that path was not pre-populated in the image builder for Endless Key content that is preloaded into the image. This is fundamentally because the image builder doesn't use the copy of Kolibri that is inside the Endless Key Flatpak to preload content: it sets up its own temporary virtualenv with Kolibri inside and uses that. But the xdg plugin was not installed to that environment.

In order to generate .desktop files at image build time, install kolibri-app-desktop-xdg-plugin into the temporary venv, and enable it. This matches the corresponding logic in 60-kolibri-content. Unfortunately this is not sufficient:

1. The generated .desktop files are named based on the FLATPAK_ID environment variable, defaulting to org.learningequality.Kolibri if it is not set. Since we are not running a Flatpak, we need to set it by hand.

2. The URI scheme used in the .desktop files is based on a variable in the kolibri_app.config module, which is installed by kolibri-gnome-app and derives its contents from options passed to meson setup. Again, since we are not using the Flatpak build, this module does not exist. Rather than building & installing kolibri-gnome, or installing a fake version of this module that contains the fields we need, munge the desktop files with `sed`.

3. The .desktop files generated by the Flatpak include a TryExec= line, which is there to ensure that the launchers are not shown if the Flatpak is removed from the system. It is derived from the contents of the /.flatpak-info file, though with the Flatpak commit ID replaced with `active`. Since we are not using the Flatpak build, this file does not exist and this logic is not triggered. Use desktop-file-edit to add the field by hand.

I'm not exactly delighted with this patch but it does work. Longer term I would hope that we can either retire Endless Key in favour of generic Kolibri, and/or make the image builder use the appropriate Flatpak rather than installing its own copy of Kolibri, which has other downsides.

https://phabricator.endlessm.com/T35760
https://phabricator.endlessm.com/T35761

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -300,6 +300,12 @@ regular_users_can_manage_content = false
 # pre-release not on PyPI.
 kolibri_pkgspec = https://github.com/learningequality/kolibri/releases/download/v0.16.0-beta6/kolibri-0.16.0b6-py2.py3-none-any.whl
 
+# Kolibri xdg desktop file plugin to use when preloading channels. Out of the
+# box these desktop files are not used for Endless Key, but some OS images may
+# add them to the search path, so we should ensure they are pregenerated in the
+# OS image.
+kolibri_app_desktop_xdg_plugin_version = 1.3.0
+
 # Which Endless Key collections to preload in the image.
 # Must match the name of one of the collections shipped with the Endless Key
 # (ex. artist, explorer, spanish etc).
--- a/hooks/image/53-ek-content-preload
+++ b/hooks/image/53-ek-content-preload
@@ -37,11 +37,13 @@ venv_dir="${EIB_TMPDIR}/kolibri-content-venv"
 python3 -m venv ${venv_dir}
 source ${venv_dir}/bin/activate
 pip install "${EIB_ENDLESSKEY_KOLIBRI_PKGSPEC}"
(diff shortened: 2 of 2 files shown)


wjt:
Test build running at https://ci.endlessos.org/job/image-build-amd64/31512

dbnicholson (approved):
Ugly, but it makes sense if it works. Way to thrash through it!

dylanmccall (approved):
:scream_cat: But yes I think this is okay, and you covered all the issues I can think of here.

wjt:
It does indeed seem to work. I've rebased this to the common ancestor of `eos6.0` and `master` to merge into both.

PR #175, opened 2025-05-07
tests: Replace GitHub Actions ubuntu-20.04 runner with 24.04

starnight:
GitHub announced:
The Ubuntu 20.04 Actions runner image will begin deprecation on 2025-02-01 and will be fully unsupported by 2025-04-15. [1]

So, bump the runner to ubuntu-latest directly.

[1]: https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,8 +14,8 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - ubuntu-20.04
           - ubuntu-22.04
+          - ubuntu-24.04
     runs-on: ${{ matrix.os }}
     name: Run tests
     steps:


.github/workflows/tests.yml
```diff
@@ -14,8 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - ubuntu-20.04
-          - ubuntu-22.04
+          - ubuntu-latest
```
wjt, line 17: I believe the purpose of having more than one runner configured was to check that the image builder can be used on the current and previous Ubuntu LTS, as a proxy for "widely-used Linux distro".

In particular, `ubuntu-22.04` has Python 3.10 https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md while `ubuntu-24.04` (which is `ubuntu-latest` at present) has Python 3.12. https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

The image builder runs itself in an Endless OS environment which has Python 3.11 at present. So if you remove `ubuntu-22.04` from the test matrix, you will remove the implicit test that no Python 3.12 features are used, and so it is more likely that we will only catch such issues post merge.

I would prefer `ubuntu-22.04` and `ubuntu-24.04`.

starnight: Python version is a concrete reason!

starnight: Updated in the new commit.

PR #178, opened 2025-09-01
Changes to build EOS7

ssssam:
Combined with the changes in https://github.com/endlessm/eos-build-meta/pull/83, this can build bootable images of the EOS7 prototype.

I have the following settings in `config/local.ini` when building images:

```
[build]
branch = eos7

[image]
# Skip GPG signing of the completed image files since the default key ID
# isn't available. Alternatively, the key ID for any private key in the
# GPG homedir at /etc/eos-image-builder/gnupg.
signing_keyid =

compression = gz

[flatpak]
enable = false

[flatpak-remote-eos-apps]
# Completely override the set of installed EOS apps for testing. This will
# take precedence over all other image variant configuration. This can
# be useful if the exact app set is known or if the build is for testing
# and the full multi-gigabyte app set is not needed.
apps =
  com.endlessm.resume

[flatpak-remote-flathub]
# Completely override the set of installed Flathub apps. See documentation
# for eos-apps above.
apps =
  com.endlessm.photos

[ostree]
deploy_server_url = http://192.168.122.1:8000
pull_server_url = http://192.168.122.1:8000
dev_repo_path =
repo =
```

https://github.com/user-attachments/assets/248cb630-12ec-4cde-8543-9570685f7b06


dbnicholson:
> > Looks good to me. Any idea why the signing key needed to be updated? The old one hadn't expired.
> 
> This was needed to import the CI builds from branch os/eos/amd64/eos7. If you use the old key, OSTree doesn't accept the commit
> 
> ```
> GPG: Verification enabled, found 2 signatures:                                                   
>                                                                                                  
>   Signature made Thu 04 Sep 2025 16:00:44 CEST using RSA key ID 9E08D8DABA02FC46
>   Can't check signature: public key not found
>                                                 
>   Signature made Thu 04 Sep 2025 16:00:44 CEST using RSA key ID FCF17B17F1F8E157
>   Can't check signature: public key not found
> ```
> 
> I suppose the new CI is signing with a different key to eos-ostree-builder ?

That's bizarre. It's the exact same key but with the expiration date extended. Neither eos-ostree-builder nor the new CI do the signing with this key at all. It's done on the server post-push. Hmm, I wonder if this is https://github.com/ostreedev/ostree/issues/3386, which ended up being a gnupg issue. We should definitely make sure this key update has been delivered to EOS 6 users before upgrading to EOS 7.

stages/eib_image
```diff
@@ -425,8 +425,43 @@ EOF
...
+        # I also copied the mokmanager binary signed with Endless' certificate
+        # from an EOS6 system. This could instead be built in eos-build-meta
+        # and signed via the Endless secure-boot signing API.
+
+        # Install the "endless" UEFI bootloader entry.
+        mkdir -p "${ESP}/EFI/endless"
+        cp "${DEPLOY}"/usr/lib/efi_binaries/EFI/endless/grubx64.efi "${ESP}/EFI/endless/grubx64.efi"
+        cp "${EIB_DATADIR}"/shim-amd64/bootx64.csv "${ESP}/EFI/endless/BOOTX64.CSV"
+        cp "${EIB_DATADIR}"/shim-amd64/mmx64.efi "${ESP}/EFI/endless/mmx64.efi"
+        cp "${EIB_DATADIR}"/shim-amd64/shimx64-signed.efi "${ESP}/EFI/endless/shimx64.efi"
```
dbnicholson, line 452: This seems OK for the short term but ultimately wrong. If users are to be able to update shim on existing systems, then shim needs to be in the ostree.

stages/eib_manifest
```diff
@@ -23,4 +23,4 @@ run_hooks manifest
 
 # Assemble all the json fragments to one file
 mkdir -p "${EIB_OUTDIR}"
-"${EIB_HELPERSDIR}"/assemble-manifest > "$(eib_outfile manifest.json)"
+#"${EIB_HELPERSDIR}"/assemble-manifest > "$(eib_outfile manifest.json)"
```
dbnicholson, line 26: The image server is dependent on this file, so you still need to generate it.

run-build
```diff
@@ -196,45 +194,12 @@ class ImageBuildRoot(object):
...
-            os.path.join(self.builddir, 'var/lib/dpkg'),
-        )
-
         # update-catalog requires /var/lib/sgml-base to exist
         os.makedirs(os.path.join(self.builddir, 'var/lib/sgml-base'))
 
         # update-xmlcatalog requires /var/lib/xml-core to exist
         os.makedirs(os.path.join(self.builddir, 'var/lib/xml-core'))
 
-        log.info('Installing buildroot packages')
```
dbnicholson, line 210: This is separate from the package metadata. Can you split it into a different commit?

stages/eib_image
```diff
@@ -155,11 +155,12 @@ if [[ "${EIB_IMAGE_SDBOOT}" = "true" ]]; then
...
 
+# FIXME: need a manifest in a new way
 # Output package list using the dpkg database from the deployment's
 # immutable copy in /usr/share/dpkg/database.
-dpkgdir="${OSTREE_DEPLOYMENT}"/usr/share/dpkg/database
-packageinfo="${EIB_TMPDIR}"/packages.txt
-dpkg-query -W --admindir="${dpkgdir}" > "${packageinfo}"
+#dpkgdir="${OSTREE_DEPLOYMENT}"/usr/share/dpkg/database
+#packageinfo="${EIB_TMPDIR}"/packages.txt
+#dpkg-query -W --admindir="${dpkgdir}" > "${packageinfo}"
```
dbnicholson, line 164: Please delete this instead of leaving commented lines that won't work anymore.

dbnicholson (changes requested):

hooks/image/50-systemd-sysusers.chroot
```diff
@@ -0,0 +1,3 @@
+# Populate /etc/passwd and /etc/group with system users configured in
+# /usr/lib/sysusers.d.
+systemd-sysusers
```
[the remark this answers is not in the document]
ssssam: There are two places we run systemd-sysusers. One is setting up the buildroot (very early), and the other place is here.

I got confused initially and only ran it in the buildroot, but it needs to happen in both places.

Agree with moving the hook earlier in any case, will update

stages/eib_image
ssssam: That will also require some mechanism for `ostree admin deploy` to get the new shim binary from wherever it is in the ostree and overwrite the ESP. As far as I know, nothing currently does that. (I may be wrong!)

stages/eib_manifest
ssssam: Hmm, ok. I was hoping we could remove it until https://github.com/endlessm/eos-build-meta/issues/76 is ready. But I could get it working again in this branch, so that we don't hit errors when uploading to the image server.

stages/eib_image
```diff
@@ -51,7 +51,7 @@ ostree --repo="${REPOPATH}" pull-local --disable-fsync \
   --remote=${EIB_OSTREE_REMOTE} ${EIB_OSTREE_REPODIR} ${EIB_OSTREE_REF_DEPLOY}
 
 # Pull the repo metadata ref so USB updates are possible without going online
-ostree --repo="${REPOPATH}" pull ${EIB_OSTREE_REMOTE} ostree-metadata
+##ostree --repo="${REPOPATH}" pull ${EIB_OSTREE_REMOTE} ostree-metadata
```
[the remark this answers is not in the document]
ssssam: I can undo this, i needed it for local builds but presumably shouldn't be removed in 'master'.

Do you have any info about what this is doing, by the way? I couldn't work it out

stages/eib_manifest
dbnicholson: Actually, I'm pretty sure the image server will survive if the file isn't included, but it's certainly desired. Are there errors from doing this or are you just worried the data will be wrong?

stages/eib_image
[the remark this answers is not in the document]
dbnicholson: Wait, now I remember - https://docs.flatpak.org/en/latest/usb-drives.html. It's using basically the same mechanism as the P2P setup where the collection ID is required. Except this isn't a flatpak remote...

Bah. Maybe the thing to do is to move the `fetch-remote-collection-id` call up and make it unconditional. I'm pretty sure it will just have empty output if the remote doesn't have a collection ID. Then surround the `ostree-metadata` fetch in another `if [ -n "$collection_id" ]` block.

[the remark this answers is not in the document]
ssssam: Interesting, I thought we were passing a collection ID even in the local development flow. I'll double check that.

ssssam:
All the comments above are dealt with in 34a89ddaf3c80f8a78c302e516241410cfa0aa0c. In particular:

  * I spent more time cleanly removing 'buildroot.packages' option instead of doing half a job
  * Shim binaries are now copied from the ostree. https://github.com/endlessm/eos-build-meta/pull/83 updated accordingly.
  
Tested alongside eos-build-meta.git commit 24ec43cd44c606212e2ce681315eae53d732e971, it produced a bootable image with the same caveats as above.

stages/eib_image
ssssam: Cleaned this up in 34a89ddaf3c80f8a78c302e516241410cfa0aa0c

run-build
ssssam: Done in 34a89ddaf3c80f8a78c302e516241410cfa0aa0c

hooks/image/50-systemd-sysusers.chroot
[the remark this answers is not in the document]
ssssam: Done in 34a89ddaf3c80f8a78c302e516241410cfa0aa0c

stages/eib_image
```diff
@@ -103,7 +103,8 @@ if [[ "${EIB_IMAGE_SDBOOT}" = "true" ]]; then
 fi
 
 kargs=(--karg=rw --karg=splash \
-    --karg=plymouth.ignore-serial-consoles --karg=quiet --karg=loglevel=0)
+    --karg=plymouth.ignore-serial-consoles --karg=quiet --karg=loglevel=0 \
+    --karg=systemd.firstboot=no)
```
dsd, line 107: I'd like to understand why this is useful/necessary compared to eos6.
eos6 also seems to have this systemd firstboot service, not sure if it runs or not.
Perhaps it doesn't because our ostree builder creates /etc/machine-id as an empty file?

stages/eib_image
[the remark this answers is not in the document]
dsd: Is this change (in a commit about shim) still pending action?

config/defaults.ini
```diff
@@ -135,7 +135,6 @@ hooks_add =
   70-flatpak-appstream-catalog
   70-flatpak-manifest
   70-ostree-manifest
-  70-packages-manifest
```
dsd, line 137: if it's now unused, please can you delete this file as well

config/defaults.ini
```diff
@@ -54,53 +54,49 @@ manifestdir = ${tmpdir}/manifest
...
-  wget
-  xz-utils
-
-# Apt archive cache directory
-aptcache_dir = ${build:contentdir}/archives
-
-# Apt archive cache maximum size in bytes (default 2 GB)
-aptcache_max_size = 2147483648
+# NOTE: In eos7, the "buildroot" doesn't have a package manager.
+# This list is preserved as a reference while we work on eos7.
```
dsd, line 58: Please just delete the lists -- as a reference we can look at eos6 branch or git history

stages/eib_image
```diff
@@ -420,8 +420,43 @@ EOF
...
+        # The eos-build-meta GRUB elements install the GRUB x86_64-efi binary
+        # to same path ready to go in the ESP.
+        #
+        # For EOS7, we copy the Shim binary signed with Microsoft's certificate
+        # from Git. It's not gonna change in the short term as we would need
+        # to request a new signature from Microsoft.
+        #
+        # I also copied the mokmanager binary signed with Endless' certificate
+        # from an EOS6 system. This could instead be built in eos-build-meta
+        # and signed via the Endless secure-boot signing API.
```
dsd, line 438: This comment seems more appropriate within the commit message adding shim to eos-build-meta.

Overall do we need to include any of these changes in the image builder here? Can we just have the ostree build set up /usr/lib/efi_binaries/EFI in the currently expected way and have the existing image builder code run?

dsd (changes requested):

ssssam:
> Is this change (in a commit about shim) still pending action?

Mysteriously Github gives me no way to reply on that comment. The commented ostree-metadata line was a mistake on my part, I have the change locally and must have accidentally commited it.

ssssam:
Updated as follows in 38c0c375df6550c5d7f95bd95ca272d91b2061da.

Cleanups requested above all done

`stages/eib_image` now has only minimal changes to the ESP setup. As explained in the commit message:

    In eos6 there would be files in `/usr/lib/efi_binaries/EFI/endless`,
    some of which were duplicated under different names in
    `/usr/lib/efi_binaries/EFI/BOOT`.
    
    In eos7, the same binaries are present but only within `EFI/endless`.
    We leave the image builder to construct the final layout of the EFI
    System Partition from those.

This is why there are some changes.

stages/eib_image
ssssam: Simplified the changes in 38c0c375df6550c5d7f95bd95ca272d91b2061da

config/defaults.ini
ssssam: Done in 38c0c375df6550c5d7f95bd95ca272d91b2061da

config/defaults.ini
ssssam: Done in 38c0c375df6550c5d7f95bd95ca272d91b2061da

stages/eib_image
ssssam: Ok. I can take a look at changing this next week.

dsd (changes requested):
Thanks - I think we're almost there!
Can you please squash "Remove package lists altogether" into "Remove the ability to install packages in the buildroot"
and "Remove unused packages manifest hook" into "Remove package metadata"

stages/eib_image
```diff
@@ -457,9 +457,8 @@ EOF
         # in the dual-boot case, and simplicity in the ISO/live case, we
         # arrange for EFI/ to have the right structure:
         mkdir -p "${boot_zip_dir}/EFI/BOOT"
+        cp "${DEPLOY}"/usr/lib/efi_binaries/EFI/endless/shimx64-signed.efi "${boot_zip_dir}/EFI/BOOT/BOOTX64.EFI"
```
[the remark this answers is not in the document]
ssssam: Ahh -- line 456 isn't the part of the script which creates the ESP, but there is indeed a need for some special handling on line 424 which I had done in commit 80d299f, then removed and it's now broken. 

Next week I will rename the file in the eos-build-meta, and do a another full test of image building + booting.

stages/eib_image
dbnicholson: FWIW, @wjt has done some deep investigation here in the past. See https://phabricator.endlessm.com/T35527. That also includes some of my "set root password on first boot" toying I was doing. That also includes another reason why GNOME OS has `systemd.firstboot=no` on the command line - it ensures that systemd-firstboot runs if the `ConditionFirstBoot` conditions are met but makes it run non-interactively.

Anyways, the conclusion of that task was to leave Endless the way it is for now.

hooks/image/50-locales
```diff
@@ -1,8 +1,17 @@
...
+import os
+from pathlib import Path
+
+manifestdir = os.environ["EIB_MANIFESTDIR"]
+locales = os.environ.get("EIB_FLATPAK_LOCALES", "").splitlines()
+locales = [loc for loc in locales if loc]
+
+outpath = Path(manifestdir) / "flatpak-locales.json"
+with open(outpath, "w") as f:
+    json.dump({"flatpak_locales": locales}, f)
```
dbnicholson, line 17: Certainly nicer!

stages/eib_image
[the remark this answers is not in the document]
dbnicholson: No, that's right the way it is. In `/EFI/BOOT` it should be named `BOOTX64.EFI` for when there are no boot options and EFI attempts to load the default name. It's `shimx64.efi` within the `/EFI/endless` directory specifically because the CSV file says that's what to use. That gets recorded in the boot option as you've shown above.

stages/eib_image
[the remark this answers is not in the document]
dbnicholson: Looks like it got removed. Would be interested to know why it wasn't working before. If you find that the local repo has a collection ID and there's no `ostree-metadata` ref after running `ostree summary --update`, then I think something would be wrong.

dbnicholson (approved):
This looks good enough to me now. The `systemd.firstboot=no` part needs to get resolved one way or the other, but I think it's fine to have it like that until it can be investigated with the machine-id file properly.

stages/eib_image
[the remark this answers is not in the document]
dsd: > No, that's right the way it is. In /EFI/BOOT it should be named BOOTX64.EFI for when there are no boot options and EFI attempts to load the default name. 

Not sure if this comment was in response to me or Sam?

I agree that /EFI/BOOT/BOOTX64.EFI is the right name for the one that gets installed there.

However I believe what is present in this proposed setup as /usr/lib/efi_binaries/EFI/endless/shimx64-signed.efi should actually be named shimx64.efi to match eos6.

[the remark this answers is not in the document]
dsd: Hmm, that commit seems to install `EFI/endless/shimx64.efi` (which is what I'd expect) but this PR tries to find it as `EFI/endless/shimx64-signed.efi`, so I think this PR does need an update. Or am I missing something?

stages/eib_image
dsd: @ssssam this is still pending - or are you thinking of doing the machine-id bit as a separate task?

ssssam: I'd rather do this separately. Opened https://github.com/endlessm/eos-build-meta/issues/95.

PR #181, opened 2025-09-22
Run systemd-tmpfiles in the ostree checkout

ssssam:
This is to fix an issue during the 50-flatpak image hook, which would fail at the appstream update with this:

    gi.repository.GLib.GError: g-io-error-quark: While pulling appstream2/x86_64 from remote flathub: open(O_TMPFILE): No such file or directory (1)

The root cause being a lack of `/var/tmp` in the ostree checkout.

It's surprising that things worked as well as they did until now, and perhaps is due to the ostree containing more things in `/var/` than it should.

### DIFF
--- a/config/defaults.ini
+++ b/config/defaults.ini
@@ -63,6 +63,7 @@ hooks_add =
 # Merged customization hooks to run.
 hooks_add =
   10-systemd-sysusers.chroot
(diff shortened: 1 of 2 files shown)


wjt (approved):
This looks good to me but I wonder if we will need to go back and clean some of these files up later in the process. Let's see!

PR #184, opened 2025-09-24
buildscript: Install awscli

starnight:
Introduce helpers/install-awscli to install awscli into the build environment. The eos-image-builder cannot publish built images to AWS s3 without awscli.

In the past (EOS 6), awscli can be installed from Debian packages repository. However, since EOS 7 based on GNOME OS, there is no package manager system. So, install awscli execution binary by following the AWS document "Installing or updating to the latest version of the AWS CLI".

Fixes: https://github.com/endlessm/eos-build-meta/issues/137

### DIFF
--- a/helpers/install-awscli
+++ b/helpers/install-awscli
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# Follow document "Installing or updating to the latest version of the AWS CLI"
+# https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
+pushd "$(mktemp -d)"
+curl "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o "awscliv2.zip"
+unzip awscliv2.zip
+./aws/install --bin-dir /usr/bin
+
+# Clean up source
+popd
+rm -r "$OLDPWD"
+
+# Show installed awscli version
+aws --version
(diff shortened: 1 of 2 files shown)


helpers/install-awscli
```diff
@@ -0,0 +1,15 @@
...
+# Follow document "Installing or updating to the latest version of the AWS CLI"
+# https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
+pushd "$(mktemp -d)"
+curl "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o "awscliv2.zip"
+unzip awscliv2.zip
+./aws/install --bin-dir /usr/bin
+
+# Clean up source
+popd
+rm -rf "$OLDPWD"
```
wjt, line 12: Why is the `-f` needed?

dbnicholson:
I just downloaded the zip file that's referenced in this file. It's python with botocore and wheel vendored.

starnight:
> This will install to /usr/local/aws-cli with a symlink in /usr/bin. Since /usr/local gets replaced in the ostree with a symlink, I think this will be broken.

Is the filesystem here the real target filesystem going to be embedded into the image?  I think it is a temporary filesystem to have a clean build environment to build the real image.

PR #186, opened 2025-10-06
Exclude codecs-extra extension

wjt:
This is freedesktop-sdk 25.08's successor to ffmpeg-full, which also includes encumbered HEIC support for libheif (which is now part of the runtime).


starnight:
Place the reference source [Rename to ffmpeg-full to codecs-extra and add to platform manifest](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/24620).

starnight (approved):
Thanks!

PR #202, opened 2026-01-09
arm64: Fix eosimpact-en image build failures

starnight:
* Here lists more Flathub apps gathered from endless-image-config, but they do not support aarch64. This avoids install the apps on arm64 images.
* EOS builds image with offline content. So, the image might be larger than 100 GB. Enable ext4's 64bit option to support larger storage.

https://app.asana.com/1/1203676861188277/project/1211949321657103/task/1212671431325376
https://app.asana.com/1/1203676861188277/project/1211949321657103/task/1212760656967809

### DIFF
--- a/stages/eib_image
+++ b/stages/eib_image
@@ -277,7 +277,6 @@ create_image() {
         bios_boot_loop=${img_loop}p2
         root_loop=${img_loop}p3
       fi
-      ext4_opts="dir_index,^huge_file"
       ESP="${ROOT}"/boot/efi
       if [[ "${EIB_IMAGE_SDBOOT}" = "true" ]]; then
         dd conv=notrunc if="${sdboot_esp_img}" of="${esp_loop}" bs=1M
@@ -286,8 +285,6 @@ create_image() {
       fi
       ;;
     *)
-      # On ARM disable 64bit ext4 option
-      ext4_opts="dir_index,^huge_file,^64bit"
       if [ "${EIB_PLATFORM:0:3}" == "rpi" ]; then
         boot_loop=${img_loop}p1
         root_loop=${img_loop}p2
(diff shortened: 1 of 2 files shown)


stages/eib_image
```diff
@@ -286,8 +286,7 @@ create_image() {
       fi
       ;;
     *)
-      # On ARM disable 64bit ext4 option
-      ext4_opts="dir_index,^huge_file,^64bit"
+      ext4_opts="dir_index,^huge_file"
```
wjt: I'm curious whether these options are needed at all now that they are the same between architectures.

You could verify this with `git blame` etc. but from memory: the reason we set the filesystem parameters at all is that at one point we created the image in multiple steps: first a minimal image containing the files used during boot (as determined by systemd-readahead); then we grew the image to its final size and copied the rest of the data. This was so that all files needed during boot would be physically close together, improving boot performance from DVD. https://blogs.gnome.org/wjjt/2017/12/06/everything-in-its-right-place/

wjt: I'm wrong. This was originally introduced specifically to disable 64bit on 32-bit ARM: commit df44a3a07ea0555c0321c50ebd245eac6e8fa80b

So you could fully revert that patch and specify `ext4_opts` in one place, rather than having this architecture-specific branching but using the same value on both branches.

wjt: I was curious why we disable `huge_file`. It dates back to @dsd's "Initial commit of new image builder" in 2014, without explanation. https://github.com/endlessm/endless-image-config/commit/0e559242743a0d410f185dd7ac88bdd07c6f54e5#diff-ca62344851add9817b4c7f08bc9a836553278fd22f36aff6c7137e251b09c70dR80

dsd: I'm not sure either! This was copying the way that OLPC images were set up from ~2009, but there isn't full history available there.
It was likely either to make the filesystem behave more efficiently for storage devices that are way smaller than 2TB, or for bootloader compatibility.
I recommend dropping this now.

starnight: Fixed as the new commits.

starnight:
Cherry picked these commits to master branch.

PR #203, opened 2026-01-13
config: Drop com.endlessm.photos

starnight:
The com.endlessm.photos is going to be EOL on Flathub [1] and has not been maintained for a long time. So, drop it from the default apps. Change the config example accordingly, too.

[1]: https://github.com/flathub/com.endlessm.photos/issues/12


starnight:
Backported to eos6.0 branch, too.

PR #206, opened 2026-08-04
eib_image: Introduce a new hook diskimage

starnight:
Introduce a new hook diskimage which will be executed before finalize a new created image. This gives the chance for disk image level modificaiton within the diskimage hooks scripts in the future.

https://app.asana.com/1/1203676861188277/project/1211949321657103/task/1216023699715178


starnight:
Cherry picked to master branch.

